PR #11, opened 2025-09-04
Add c2pa-types + typed activeManifest and manifestStore APIs

emensch:
- Add c2pa-types package with logic to automatically generate TS types from c2pa-rs using schemars and json-schema-to-typescript.
- Add activeManifest and manifestStore methods to the reader, taking advantage of these shiny new types for their output.


gpeacock (commented):
It looks ok to me, but I'm fuzzy on how you deal with resources, how do you read a thumbnail?

tmathern (approved):

PR #28, opened 2025-09-22
fix(c2pa-web): change the unregister token passed to FinalizationRegistry.register

antonpodolsky:
Running `c2pa-web` in `FF 143.0.1` throws the following error:
```
TypeError: Invalid unregister token passed to FinalizationRegistry.register
```

Change the unregister token to be the reader reference itself, instead of a `Symbol`.
This should work because the unregister token is only weakly-referenced by the registry, as per [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry).

### DIFF
--- a/.changeset/chilly-schools-behave.md
+++ b/.changeset/chilly-schools-behave.md
@@ -1,0 +1,5 @@
+---
+'@contentauth/c2pa-web': patch
+---
+
+Fix issue which crashes reader.fromBlob() and reader.fromBlobFragment() in Firefox.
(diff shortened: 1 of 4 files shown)


emensch:
I'll look into the CI failures. Looks like some flakiness in the setup.

emensch:
Hi @antonpodolsky if you want to incorporate the changes I've just merged to main into your branch that _should_ fix the CI problems. Thanks :)

PR #60, opened 2025-12-09
feat: addIngredient

cdmurph32:
This is needed for copying ingredients from one builder to another.

### DIFF
--- a/packages/c2pa-web/src/lib/builder.ts
+++ b/packages/c2pa-web/src/lib/builder.ts
@@ -88,7 +88,7 @@ export interface Builder {
   /**
    * Add an ingredient to the builder from a definition only.
    *
-   * @param ingredientDefinition Ingredient definition.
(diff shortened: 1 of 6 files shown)


packages/c2pa-web/src/lib/builder.ts
```diff
@@ -85,6 +85,13 @@ export interface Builder {
    */
   setThumbnailFromBlob: (format: string, blob: Blob) => Promise<void>;
 
+  /**
+   * Add an ingredient to the builder from a definition only.
+   *
+   * @param ingredientDefinition Ingredient definition.
```
emensch: Could you replace `Ingredient` with `{@link Ingredient}` here? That will allow typedoc to resolve this to the ingredient type and create an interactive link on the docs site.

emensch (approved):

PR #67, opened 2026-01-08
feat: Pass settings in the reader and builder constructor

cdmurph32:
Web JS version of https://github.com/contentauth/c2pa-node-v2/pull/40

### DIFF
--- a/packages/c2pa-web/src/lib/builder.spec.ts
+++ b/packages/c2pa-web/src/lib/builder.spec.ts
@@ -10,6 +10,7 @@
 import { test, describe, expect } from 'test/methods.js';
 import { ManifestDefinition, Ingredient } from '@contentauth/c2pa-types';
 import { getBlobForAsset } from 'test/utils.js';
+import { SettingsContext } from './settings.js';
 import C_JPG from 'test/assets/C.jpg';
 
 describe('builder', () => {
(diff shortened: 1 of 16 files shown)


packages/c2pa-web/src/lib/settings.ts
```diff
@@ -9,8 +9,51 @@
 
 import { merge } from 'ts-deepmerge';
 
+/**
+ * Context configuration for C2PA operations.
+ *
+ * Context encapsulates settings and configuration options for Reader and Builder operations.
+ * It provides a flexible way to configure SDK behavior including verification settings,
```
tmathern:
```suggestion
 * It provides a flexible way to configure SDK behavior including verification configuration,
```
(I find that wording less confusing, just an opinion)

packages/c2pa-web/test/manifests/C_with_CAWG_data_untrusted.ts
```diff
@@ -187,5 +186,5 @@ export default {
       ],
     },
   },
-  validation_state: 'Invalid',
+  validation_state: 'Valid',
```
tmathern, line 189: Heads-up that that may switch back to invalid in the future. Discussions in progress here...

tmathern (approved):

tmathern:
LGTM!

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -127,6 +127,97 @@ describe('builder', () => {
             instance_id: 'ingredient-instance-123',
           });
         });
+
+        test('should read builder archive with context settings', async ({
+          c2pa,
+        }) => {
+          // Import the context serialization function
+          const { contextToWasmJson } = await import('./settings.js');
```
emensch: Just to keep style consistent I would move this to a top-level import with everything else in this file.

packages/c2pa-web/src/lib/builder.ts
```diff
@@ -22,25 +22,28 @@ import type {
 export interface BuilderFactory {
   /**
    * Create a {@link Builder} with a minimal manifest definition as its initial state.
+   * @param contextJson Optional JSON string representing context settings for the builder.
    * @returns A {@link Builder} object.
    */
-  new: () => Promise<Builder>;
+  new: (contextJson?: string) => Promise<Builder>;
```
emensch: I don't think we should force consumers to call an additional function (`contextToWasmJson`) for these APIs. We should accept an object w/ a suitable type and serialize it under the hood, mirroring what we do elsewhere: https://github.com/contentauth/c2pa-js/blob/main/packages/c2pa-web/src/lib/builder.ts#L188-L190

packages/c2pa-web/src/lib/settings.ts
```diff
@@ -94,16 +137,40 @@ type SettingsObjectType = {
...
+  await Promise.all(resolvePromises);
+
+  return JSON.stringify(snakeCaseify(mergedContext as SettingsObjectType));
+}
+
 /**
  * Resolves any trust list URLs and serializes the resulting object into a JSON string of the structure expected by c2pa-rs.
  *
  * @param settings
+ * @deprecated Use {@link contextToWasmJson} instead.
```
emensch, line 173: I see you made this part of the public API in this PR - why?

emensch (changes requested):
See comments

packages/c2pa-web/src/lib/builder.ts
cdmurph32: Yes. That isn't great. I'm still feeling my way around this SDK.

packages/c2pa-web/src/lib/settings.ts
cdmurph32: It shouldn't be. Redoing this.

packages/c2pa-web/src/lib/builder.spec.ts
cdmurph32: Getting rid of contextToWasmJson anyway, but good to know.

emensch (approved):

PR #72, opened 2026-01-22
feat: add settings to additional methods

cdmurph32:
Incorporate upstream changes
contentauth/c2pa-rs#1710

### DIFF
--- a/packages/c2pa-wasm/src/settings.rs
+++ b/packages/c2pa-wasm/src/settings.rs
@@ -10,7 +10,8 @@ use wasm_bindgen::prelude::*;
 /// Accepts a JSON-serialized string to be loaded as c2pa-rs settings.
 #[wasm_bindgen(js_name = loadSettings)]
 pub fn load_settings(settings: &str) -> Result<(), js_sys::Error> {
-    c2pa::settings::load_settings_from_str(settings, "json").unwrap();
+    c2pa::settings::Settings::from_string(settings, "json")
+        .map_err(|e| js_sys::Error::new(&format!("Failed to load settings: {}", e)))?;
 
     Ok(())
 }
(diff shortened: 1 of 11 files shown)


packages/c2pa-wasm/src/settings.rs
```diff
@@ -10,7 +10,8 @@ use wasm_bindgen::prelude::*;
 /// Accepts a JSON-serialized string to be loaded as c2pa-rs settings.
 #[wasm_bindgen(js_name = loadSettings)]
 pub fn load_settings(settings: &str) -> Result<(), js_sys::Error> {
-    c2pa::settings::load_settings_from_str(settings, "json").unwrap();
+    c2pa::settings::Settings::from_string(settings, "json")
+        .map_err(|e| js_sys::Error::new(&format!("Failed to load settings: {}", e)))?;
```
emensch: I think we can use the existing[ `error`](https://github.com/contentauth/c2pa-js/blob/main/packages/c2pa-wasm/src/error.rs) type here and just call `WasmError::from`.

packages/c2pa-wasm/src/wasm_builder.rs
```diff
@@ -92,12 +92,19 @@ impl WasmBuilder {
...
+        let builder = if let Some(ctx_json) = context_json {
+            let context = Context::new()
+                .with_settings(ctx_json.as_str())
+                .map_err(WasmError::from)?;
+            Builder::from_context(context)
+                .with_archive(stream)
+                .map_err(WasmError::from)?
+        } else {
+            Builder::from_archive(stream).map_err(WasmError::from)?
+        };
```
emensch, line 107: I'd like to eliminate or reduce the duplication here - can we brainstorm on ways that might be possible? I see this pattern repeated with small variations across all the "constructors"

emensch (approved):

PR #75, opened 2026-02-13
feat: Add support for arbitrary key/value pairs in assertion metadata

ale-adobe:
- Add support for arbitrary key/value pairs in assertion metadata in the JS SDK by updating the c2pa dependency to version 0.75.19 (from 0.75.6), which contains the relevant changes in the Rust SDK, added in https://github.com/contentauth/c2pa-rs/pull/1782. 
- Bump the wasm-bindgen dependency to latest version (0.2.108).
- Update some READMEs to add more clarity about setting up and working with this monorepo.


packages/c2pa-web/src/lib/settings.ts
```diff
@@ -151,7 +151,7 @@ const DEFAULT_SETTINGS: SettingsContext = {
 export async function contextToWasmJson(context: SettingsContext) {
   const mergedContext: SettingsContext = merge(DEFAULT_SETTINGS, context);
 
-  const resolvePromises: Promise<any>[] = [];
+  const resolvePromises: Promise<void>[] = [];
```
ale-adobe, line 154: This came up on the linter. There were other instances of `any` as well, but I felt like these could be safely changed to `void`.

README.md
```diff
@@ -7,19 +7,65 @@ A collection of libraries and tools that enable interaction with [C2PA metadata]
...
+#### Node.js
+
+Node.js **v22 or later** is required (earlier versions may fail with build errors related to unknown file extensions).
+
+If you use [nvm](https://github.com/nvm-sh/nvm), you can install the required version with:
+
+```sh
+nvm install <version>
+nvm install 22.22.0
+```
```
ale-adobe, line 25: I'm not sure that it has to be specifically v22, but at the very least, v18 was not sufficient. There may be a lower minimum version that would work.

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -11,7 +11,8 @@ import { test, describe, expect } from 'test/methods.js';
 import { ManifestDefinition, Ingredient } from '@contentauth/c2pa-types';
 import { getBlobForAsset } from 'test/utils.js';
 import { SettingsContext } from './settings.js';
-import C_JPG from 'test/assets/C.jpg';
+
+const JPEG_TEST_ASSET_PATH = 'test/assets/C.jpg';
```
emensch: Just curious - is this necessary?

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -103,14 +104,16 @@ describe('builder', () => {
...
             title: 'C.jpg',
-            format: 'image/jpeg',
+            format: blobType,
             instance_id: 'ingredient-instance-123',
           };
-
-          const blob = await getBlobForAsset(C_JPG);
-          await builder.addIngredientFromBlob(ingredient, 'image/png', blob);
+          
+          await builder.addIngredientFromBlob(ingredient, blob.type, blob);
```
emensch: `blob.type` -> `blobType`?

ale-adobe: Oops, missed that. Thanks!

packages/c2pa-web/src/lib/builder.spec.ts
ale-adobe: No, but it felt more natural to me to define the static string in one place as a `const` for the file path. Is the existing way of using `import` more idiomatic? I hadn't seen that before.

README.md
tmathern: We should to aim for current - 2 version *tops*.

packages/c2pa-wasm/README.md
```diff
@@ -22,6 +22,12 @@ Then, all of the following prerequisites must be installed before `c2pa-wasm` ca
 
 Minimum supported Rust version: **1.86.0**.
```
tmathern: Unrelated to these changes: SHouldn't this match also the underlying Rust-SDK: https://github.com/contentauth/c2pa-rs/blob/main/sdk/Cargo.toml#L23? Or does that rust min version depend only on neon/wasm? @cdmurph32 @emensch

emensch (approved):

packages/c2pa-wasm/README.md
emensch: I believe that's correct - let's update this

packages/c2pa-web/src/lib/builder.spec.ts
emensch: Np at all - I have a preference to keep the name descriptive (`C_jpg`), but this is fine with me if you prefer.

ale-adobe: Reverted the change around this variable.

packages/c2pa-wasm/README.md
ale-adobe: Updated `README.md` to say 1.88.0, and also added it as a workspace dependency. I think this will make `cargo` commands complain if the active Rust toolchain isn't using at least version 1.88.0.

packages/c2pa-wasm/Cargo.toml
```diff
@@ -2,7 +2,8 @@
 name = "c2pa-wasm"
 version = "0.1.0"
 authors = ["Eli Mensch <mensch@adobe.com>"]
-edition = "2021"
+edition = "2024"
```
ale-adobe, line 5: Updating the edition to match the one in the `c2pa-types` package. Apparently, 2024 equates to Rust version 1.85.0 or later.

PR #77, opened 2026-02-15
feat: settings interface cleanup

emensch:


packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -127,20 +129,11 @@ describe('builder', () => {
...
 
-        test('should read builder archive with context settings', async ({
-          c2pa,
-        }) => {
-          // Configure builder to generate C2PA archive
-          const builderContext: SettingsContext = {
-            builder: {
-              generateC2paArchive: true,
-            },
-          };
```
emensch, line 142: @cdmurph32 as this is now properly included in defaults, it wasn't doing anything

packages/c2pa-web/src/lib/settings.ts
```diff
@@ -146,31 +128,7 @@ const DEFAULT_SETTINGS: SettingsContext = {
...
-  }
-
-  if (mergedContext.cawgTrust) {
-    resolvePromises.push(resolveTrustSettings(mergedContext.cawgTrust));
-  }
-
-  await Promise.all(resolvePromises);
-
-  return JSON.stringify(snakeCaseify(mergedContext as SettingsObjectType));
-}
```
emensch, line 167: Identical behavior and interface to `settingsToWasmJson`. Felt reasonable to just standardize around the `Settings` interface and a single function.

cdmurph32 (approved):
Approved

packages/c2pa-web/src/lib/settings.ts
```diff
@@ -50,24 +50,6 @@ export interface SettingsContext {
   builder?: BuilderSettings;
 }
 
-/**
- * Settings used to configure the SDK's behavior.
- *
- * @deprecated Use {@link SettingsContext} instead. Settings will be removed in a future version.
```
ale-adobe, line 56: 🎉

packages/c2pa-web/src/lib/settings.ts
```diff
@@ -146,31 +128,7 @@ const DEFAULT_SETTINGS: SettingsContext = {
...
-
-  return JSON.stringify(snakeCaseify(mergedContext as SettingsObjectType));
-}
-
-/**
- * Resolves any trust list URLs and serializes the resulting object into a JSON string of the structure expected by c2pa-rs.
- *
- * @param settings
- * @deprecated Use {@link contextToWasmJson} instead.
+ * @param settings Settings configuration object
```
ale-adobe, line 119: Existing code, but wondering if we're missing an `@returns` annotation here, just to keep up the good hygiene. (I know it would be a little redundant given the existing description above.)

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -326,23 +309,23 @@ describe('builder', () => {
...
+            },
           };
 
           await builder.addIngredient(ingredient);
 
           const definition = await builder.getDefinition();
 
           expect(definition.ingredients).toHaveLength(1);
-          expect(definition.ingredients[0]).toMatchObject(ingredient);
+          expect(definition.ingredients?.[0]).toMatchObject(ingredient);
```
ale-adobe: Thanks for adding these; I was considering doing the same in my last PR.

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -62,13 +62,13 @@ describe('builder', () => {
             claim_generator_info: [
               {
                 name: 'c2pa-web-test',
-                version: '1.0.0'
+                version: '1.0.0',
```
ale-adobe: Do we prefer to leave the trailing commas?

packages/c2pa-web/src/lib/settings.ts
```diff
@@ -257,3 +214,20 @@ const containsCerts = (content: string): boolean =>
...
+
+async function retrieveFromCacheOrFetch(url: string): Promise<string> {
+  const maybeCached = cache.get(url);
+
+  if (maybeCached) {
+    return maybeCached;
+  }
+
+  const res = await fetch(url);
+  const text = await res.text();
```
ale-adobe, line 208: How big are these objects, and how many are we expecting to realistically cache in practice? Just wondering how much memory we could be consuming with this.

packages/c2pa-web/src/lib/settings.spec.ts
```diff
@@ -119,16 +129,19 @@ describe('settings', () => {
...
           )
         );
 
         const settingsString = await settingsToWasmJson({
           trust: {
-            userAnchors: ['http://userAnchors', 'http://userAnchors'],
+            userAnchors: [
+              'http://userAnchorsConcat',
+              'http://userAnchorsConcat',
+            ],
```
ale-adobe, line 134: Could you explain why we list the same URL twice in this setting?

packages/c2pa-web/src/lib/settings.spec.ts
```diff
@@ -8,8 +8,18 @@
  */
 
 import { test, describe, expect } from 'test/methods.js';
-import { settingsToWasmJson } from './settings.js';
 import { http, HttpResponse } from 'msw';
+import { settingsToWasmJson } from './settings.js';
+
+/**
+ * NOTE: settingsToWasmJson relies on a module-level HTTP cache that I could not
```
ale-adobe: Are you referring to the in-memory cache defined [here](https://github.com/contentauth/c2pa-js/pull/77/changes#diff-f22b8b1c217de5b00c7d7e3b15de63db773979478821a0794606fa4a3918c79fR218)? 🤔

packages/c2pa-web/src/lib/reader.spec.ts
```diff
@@ -235,6 +235,43 @@ describe('reader', () => {
     c2pa.dispose();
   });
 
+  test('should use local "context" settings when provided', async () => {
```
ale-adobe: Should there be a similar test to verify the override behavior for `builder`?

packages/c2pa-web/src/lib/reader.spec.ts
```diff
@@ -235,6 +235,43 @@ describe('reader', () => {
...
+    const settings: Settings = {
+      verify: {
+        verifyTrust: false,
+      },
+      cawgTrust: {
+        verifyTrustList: false,
+      },
+    };
+
+    const contextSettings: Settings = {
```
ale-adobe: For clarity when reading the test, maybe rename to something like `overrideSettings`?

ale-adobe (commented):
Looks good overall, but leaving a bunch of questions for the sake of my own understanding. 😄

packages/c2pa-web/src/lib/builder.spec.ts
emensch: @ale-adobe I have no preference, but your comment here and [here](https://github.com/contentauth/c2pa-js/pull/77#discussion_r2818715079) I think suggest a need for some pre-commit lint and formatting hooks. 

We could set up [lefthook](https://github.com/evilmartians/lefthook) (or another commit hook system) 🤔

packages/c2pa-web/src/lib/settings.ts
emensch: It's a good question. I think the trust config that we load for Inspect amounts to about 70kb.

It also has me wondering if maybe I'm going a little too ham here - perhaps it's enough to just rely on browser HTTP caching? Still, it feels a little ugly to let the requests through...

packages/c2pa-web/src/lib/settings.spec.ts
emensch: This tests that an array of URLs will be requested and concatenated correctly. We could use different URLs, but then we have to set up a second mock. It seems adequate to just use the same URL twice.

Open to being convinced otherwise!

packages/c2pa-web/src/lib/settings.spec.ts
emensch: Yup, that's the one

packages/c2pa-web/src/lib/reader.spec.ts
emensch: Yes, good call.

packages/c2pa-web/src/lib/settings.ts
ale-adobe: I'm inclined to think the browser cache would be sufficient, unless we have proof that it's not good enough otherwise.

packages/c2pa-web/src/lib/settings.spec.ts
ale-adobe: Sounds fine to me, just wanted to check!

packages/c2pa-web/src/lib/settings.ts
```diff
@@ -116,63 +98,28 @@ export interface VerifySettings {
...
-   */
-  thumbnail?: BuilderThumbnailSettings;
-}
-
-export interface BuilderThumbnailSettings {
-  /*
-   * Whether or not to automatically generate thumbnails.
-   * The default value is true.
-   */
-  enabled: boolean;
```
emensch, line 133: Spoke to @cdmurph32, we agreed that thumbnail settings don't do anything unless c2pa-rs is compiled with the `add_thumbnails` feature, which we do not currently use for c2pa-wasm.

ale-adobe (approved):
LGTM, thanks for the cleanup!

PR #80, opened 2026-02-19
style: Add ESLint rule comma-dangle to never be allowed at error level

ale-adobe:
Add `comma-dangle` ESLint rule to trigger an error when detected. This will fail CI checks, forcing developers to run the linter and commit the fixes before proceeding. They can run the linter using `nx lint <package>`.

I also took this opportunity to run the linter on the `c2pa-web` and `c2pa-wasm` packages, which is why there are style changes across those packages.


packages/c2pa-web/eslint.config.mjs
```diff
@@ -11,10 +11,11 @@ import baseConfig from '../../eslint.config.mjs';
 export default [
   ...baseConfig,
   {
-    files: ['test/**', '**/*.spec.ts'],
+    files: ['src/**/*.ts', 'test/**', '**/*.spec.ts'],
```
ale-adobe, line 14: Adding all TS files in the `src/` subdirectory.

packages/c2pa-web/eslint.config.mjs
```diff
@@ -11,10 +11,11 @@ import baseConfig from '../../eslint.config.mjs';
...
     rules: {
       'no-empty-pattern': 'off',
-      '@typescript-eslint/no-non-null-assertion': 'off'
+      '@typescript-eslint/no-non-null-assertion': 'off',
+      'comma-dangle': ['error', 'never']
```
ale-adobe, line 18: If detected, will produce an **error**, and is **never** allowed. This will cause the CI check step `pnpm ci:check` to fail if a dangling comma is encountered.

ale-adobe (commented):
Some comments for extra context.

emensch (approved):

PR #88, opened 2026-02-27
fix: broken error handling in Firefox

emensch:
Due to some mysterious behavior (likely in wasm-bindgen), `JsError`s were not being thrown properly by wasm-bindgen generated functions. 

As a workaround, we instead report our errors as `JsString`s, which we then "convert" into Error objects at the boundary between JS/WASM.

I certainly do not love this as a fix. I am hopeful that the underlying issue in wasm-bindgen will eventually be fixed. See: https://github.com/wasm-bindgen/wasm-bindgen/issues/4961

Additionally, this PR adds Firefox and Webkit to the test runner.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,3 +1,4 @@
 {
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": "none"
 }
--- a/packages/c2pa-web/src/lib/worker.ts
+++ b/packages/c2pa-web/src/lib/worker.ts
@@ -203,7 +203,7 @@ function wrapFunctions<T extends Record<string, (...args: any[]) => any>>(
 ): T {
   const wrappedFunctions = {} as Record<string, (...args: any[]) => any>;
 
-  for (let [fnName, fn] of Object.entries(functions)) {
+  for (const [fnName, fn] of Object.entries(functions)) {
     wrappedFunctions[fnName] = async (...args: any[]) => {
       try {
         return await fn(...args);
--- a/packages/c2pa-web/src/lib/worker.ts
+++ b/packages/c2pa-web/src/lib/worker.ts
@@ -25,7 +25,7 @@ const builderMap = createWorkerObjectMap<WasmBuilder>();
 const tx = createWorkerTx();
 
 rx(
-  wrapFunctions({
+  wrapFunctionsForErrorHandling({
     async initWorker(module, settings) {
       initSync({ module });
       if (settings) {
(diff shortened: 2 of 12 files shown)


packages/c2pa-web/vite.config.ts
```diff
@@ -69,7 +62,11 @@ export default defineConfig(() => ({
       enabled: true,
       provider: 'playwright',
       // https://vitest.dev/guide/browser/playwright
-      instances: [{ browser: 'chromium' }],
+      instances: [
+        { browser: 'chromium' },
+        { browser: 'firefox' },
+        { browser: 'webkit' }
```
ale-adobe, line 68: 💯

.prettierrc
```diff
@@ -1,3 +1,4 @@
 {
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": "none"
```
ale-adobe, line 3: I think the linter should do this already, but I suppose it doesn't hurt to also have it configured for Prettier.

packages/c2pa-web/src/lib/worker.ts
```diff
@@ -24,166 +24,198 @@ const builderMap = createWorkerObjectMap<WasmBuilder>();
...
+
+/**
+ * Wraps all functions with additional error-handling code that converts any thrown strings into Error objects.
+ * This is only necessary because a bug (likely in wasm-bindgen, see https://github.com/wasm-bindgen/wasm-bindgen/issues/4961)
+ * prevents the proper handling of Error objects. As a workaround, we throw strings from our wasm-bindgen
+ * functions and convert them into errors here.
+ *
+ * I hope that one day this can be removed :)
+ */
+function wrapFunctions<T extends Record<string, (...args: any[]) => any>>(
```
ale-adobe: nit: alternate naming suggestion
```suggestion
function wrapFunctionsForErrorHandling<T extends Record<string, (...args: any[]) => any>>(
```

packages/c2pa-web/src/lib/worker.ts
```diff
@@ -24,166 +24,198 @@ const builderMap = createWorkerObjectMap<WasmBuilder>();
...
+  })
+);
+
+/**
+ * Wraps all functions with additional error-handling code that converts any thrown strings into Error objects.
+ * This is only necessary because a bug (likely in wasm-bindgen, see https://github.com/wasm-bindgen/wasm-bindgen/issues/4961)
+ * prevents the proper handling of Error objects. As a workaround, we throw strings from our wasm-bindgen
+ * functions and convert them into errors here.
+ *
+ * I hope that one day this can be removed :)
```
ale-adobe: nit: Prefer to leave commentary like this out of the actual function docs.

packages/c2pa-web/src/lib/worker.ts
```diff
@@ -24,166 +24,198 @@ const builderMap = createWorkerObjectMap<WasmBuilder>();
...
+      builder.free();
+      builderMap.remove(builderId);
+    }
+  })
+);
+
+/**
+ * Wraps all functions with additional error-handling code that converts any thrown strings into Error objects.
+ * This is only necessary because a bug (likely in wasm-bindgen, see https://github.com/wasm-bindgen/wasm-bindgen/issues/4961)
+ * prevents the proper handling of Error objects. As a workaround, we throw strings from our wasm-bindgen
```
ale-adobe:
```suggestion
 * prevents the proper handling of Error objects. As a workaround, we "throw" strings from our wasm-bindgen
```

ale-adobe (approved):
Not a Rust expert, so won't comment on any language-specific things. Overall short-term workaround fix looks good to me. Have some non-blocking comments and suggestions. Thanks for the fix!

packages/c2pa-web/src/lib/worker.ts
emensch: :(

ale-adobe (approved):
Firefox skip LGTM. Thanks!

PR #95, opened 2026-03-21
fix: adding ingredients from blobs should fetch remote manifests

cdmurph32:
Builder.addIngredientFromBlob needs to call Builder.add_ingredient_from_stream_async() in the Rust SDK. The sync version doesn't fetch remote manifests.


ale-adobe (approved):
LGTM, thanks for the fix.

PR #100, opened 2026-04-21
fix: fix possible offset overflow

cdmurph32:


packages/c2pa-wasm/src/stream/blob_stream.rs
```diff
@@ -163,6 +176,27 @@ mod tests {
         assert_eq!(bytes_read, 0)
     }
 
+    #[wasm_bindgen_test]
+    fn test_crafted_png_seek_does_not_wrap() {
+        // Regression test for CAI-11202 / VULN-34123:
```
emensch: The tiniest of nits - would prefer we don't reference internal jira tickets in open-source code.

emensch (approved):

.changeset/tall-crews-poke.md
```diff
@@ -0,0 +1,6 @@
+---
+'@contentauth/c2pa-wasm': patch
+'@contentauth/c2pa-web': patch
+---
+
+fix possible blobstore offset overflow
```
emensch: `blobstore` -> `BlobStream`?

cdmurph32: 🤦

packages/c2pa-wasm/src/stream/blob_stream.rs
cdmurph32: Yeah. That was dumb.

PR #105, opened 2026-05-04
feat: Add redaction logic

cdmurph32:
Import C2paReason type from https://github.com/contentauth/c2pa-rs/pull/2100

### DIFF
--- a/packages/c2pa-web/src/lib/builder.ts
+++ b/packages/c2pa-web/src/lib/builder.ts
@@ -12,6 +12,7 @@ import { getSerializablePayload, type Signer } from './signer.js';
 import type {
   Action,
   BuilderIntent,
+  C2paReason,
   Ingredient,
   ManifestDefinition
 } from '@contentauth/c2pa-types';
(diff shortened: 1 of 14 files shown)


packages/c2pa-web/src/lib/builder.ts
```diff
@@ -92,6 +93,17 @@ export interface Builder {
    */
   setThumbnailFromBlob: (format: string, blob: Blob) => Promise<void>;
 
+  /**
+   * Redact an assertion from an ingredient manifest.
```
tmathern, line 97: Or a thumbnail? Does this handle redacting thumbnails too (it should, since they have/are URI, but gathering them is different than gathering action assertions).
It may be worth adding one test that redacts a thumbnail, to verify there are no surprises here.
(The challenging ones in native SDKs were to redact the thumbnail on an ingredient added as ingredient, to redact the thumbnail add_ingredient adds when the Rust feature to autogenerate thumbnails is here).

tmathern (approved):

packages/c2pa-web/src/lib/builder.ts
cdmurph32: Sounds like a good idea.

PR #108, opened 2026-05-14
fix: Handle null value cases in snakeCaseify to avoid crashes

ale-adobe:
Avoid crash via `TypeError` by adding a null check in `snakeCaseify()`.

### DIFF
--- a/packages/c2pa-web/src/lib/settings.ts
+++ b/packages/c2pa-web/src/lib/settings.ts
@@ -141,7 +141,7 @@ function snakeCaseify(object: SettingsObjectType): SettingsObjectType {
   const formattedObject = Object.entries(object).reduce(
     (formattedObject, [key, val]) => {
       formattedObject[snakeCase(key)] =
-        typeof val === 'object' ? snakeCaseify(val) : val;
+        typeof val === 'object' && val !== null && !Array.isArray(val) ? snakeCaseify(val) : val;
       return formattedObject;
     },
     {} as SettingsObjectType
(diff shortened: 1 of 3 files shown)


packages/c2pa-web/src/lib/settings.ts
```diff
@@ -141,7 +141,7 @@ function snakeCaseify(object: SettingsObjectType): SettingsObjectType {
   const formattedObject = Object.entries(object).reduce(
     (formattedObject, [key, val]) => {
       formattedObject[snakeCase(key)] =
-        typeof val === 'object' ? snakeCaseify(val) : val;
+        typeof val === 'object' && val !== null ? snakeCaseify(val) : val;
```
tmathern, line 144: Is undefined working OK? That's likely another one that may crash this

emensch: I think that should be _ok_ because it won't pass the `typeof val === 'object'` check, which `null` bizarrely does.

emensch (approved):

packages/c2pa-web/src/lib/settings.ts
ale-adobe: Yeah, weird TypeScript things. 🫠

tmathern (approved):

packages/c2pa-web/src/lib/settings.ts
tmathern: Right. I was thinking in js due to the repo name.

PR #109, opened 2026-05-14
fix: Add checks for valid trust keys and fetched resource response size

ale-adobe:
This PR adds the following guards and checks to the `resolveTrustSettings` API:
- Only process keys in the `TrustSettings` interface when resolving the passed-in trust settings (other keys could exist, such as those in an extending interface like `CawgTrustSettings`), which are known to be the ones that we want to actually process.
- Cap responses from fetching resources from trust URLs to 1MB, which should be a generous upper bound on what a realistic trust anchor or config file could reach. Anything over that size will throw an error.

### DIFF
--- a/packages/c2pa-web/src/lib/settings.ts
+++ b/packages/c2pa-web/src/lib/settings.ts
@@ -77,6 +77,14 @@ export interface TrustSettings {
   allowedList?: string | string[];
 }
 
+const TRUST_SETTINGS_KEY_MAP: Record<keyof TrustSettings, true> = {
+  userAnchors: true,
+  trustAnchors: true,
+  trustConfig: true,
+  allowedList: true
+};
+const TRUST_SETTINGS_KEYS = Object.keys(TRUST_SETTINGS_KEY_MAP) as (keyof TrustSettings)[];
+
 export interface CawgTrustSettings extends TrustSettings {
   /**
    * Enable CAWG trust validation. The default value is "true."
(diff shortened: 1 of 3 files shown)


packages/c2pa-web/src/lib/settings.ts
```diff
@@ -161,9 +172,16 @@ function snakeCase(str: string): string {
...
 async function resolveTrustSettings(settings: TrustSettings): Promise<void> {
   try {
-    const promises = Object.entries(settings).map(async ([key, val]) => {
-      if (Array.isArray(val)) {
-        const promises = val.map(async (val) => {
+    const promises = Object.entries(settings)
+    .filter(([key]) => TRUST_SETTINGS_KEYS.includes(key as keyof TrustSettings))
+    .map(async ([key, val]) => {
+      if (val && typeof val === 'object' && Array.isArray(val)) {
+        // Only fetch the first MAX_URLS_PER_TRUST_SETTING URLs to prevent excessive resource consumption.
```
tmathern: How would we know it may not be enough? Bug report to whoever it happens?

ale-adobe: That's what I'm thinking... I'm not sure if there's a better way, but I also don't feel comfortable *not* implementing some sort of limit.

tmathern: I don't think we have settings or env vars for this.
@emensch @cdmurph32 may know better.
Provisional approval until they comment back!

tmathern (approved):

packages/c2pa-web/src/lib/settings.ts
emensch: I think I would consider leaving the number of URLs uncapped? My feeling is basically this: if an _implementer_ of the SDK wants to request 100s of trust lists, is it the SDK's responsibility to prevent that? It would be different if this was expected to handle arbitrary user input, of course, but I don't think that will ever be the case with settings.

Not a strong feeling, just some food for thought

emensch (approved):

packages/c2pa-web/src/lib/settings.ts
ale-adobe: After further consideration, I agree and I'll remove capping the number of URLs. It's up to the SDK consumer to decide what they pass in for their custom trust lists.

PR #112, opened 2026-05-26
docs: Add testbed readme

### LINKED ISSUE #98: c2pa-js testbed needs some docs
Re https://github.com/contentauth/c2pa-js/tree/main/tools/testbed

Initially, I thought the app was broken, because there are no instructions on how to run it. But I finally figured it out. I think there should be a short README that just explains how to run the app.

The app in `tools/testbed` is supposed to be a minimal website that can be used to view c2pa-web output and as a sandbox for local development.

If I run it using a local web server, the page displays properly without errors, but after dragging an image onto the page as instructed, the browser console displays this error:
```java
... Unsafe attempt to load URL 
<file:///Users/randmckinney/Desktop/CAI/examples/fungus*Ps*signed.jpg> from frame with URL 
<file:///Users/randmckinney/Desktop/CAI/examples/fungus*Ps*signed.jpg>. 'file:' 
URLs are treated as unique security origins.
```
If I load the page directly from the file system, the console shows a bunch of errors and the page doesn't accept the drag/dropped image.

crandmck:
Adds README for the testbed app, which may be helpful for users getting started.

Closes #98


ale-adobe (approved):
LGTM, thank you!

PR #114, opened 2026-05-30
feat: Improve error handling when fetching trust anchors

ale-adobe:
- Includes the error message directly when throwing the `Error` object in `resolveTrustSettings`
- Handle failed `fetch` calls when fetching a resource by wrapping the call in a try-catch block and reporting different error messages based on what actually happened. Network errors can be differentiated from actual errors coming from the fetch (e.g. 429, 404, etc.).

### DIFF
--- a/packages/c2pa-web/src/lib/settings.ts
+++ b/packages/c2pa-web/src/lib/settings.ts
@@ -207,7 +207,8 @@ async function resolveTrustSettings(settings: TrustSettings): Promise<void> {
 
     await Promise.all(promises);
   } catch (e) {
-    throw new Error('Failed to resolve trust settings.', { cause: e });
+    const message = e instanceof Error ? e.message : String(e);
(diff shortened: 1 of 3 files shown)


packages/c2pa-web/src/lib/settings.ts
```diff
@@ -207,7 +207,8 @@ async function resolveTrustSettings(settings: TrustSettings): Promise<void> {
 
     await Promise.all(promises);
   } catch (e) {
-    throw new Error('Failed to resolve trust settings.', { cause: e });
+    const message = e instanceof Error ? e.message : String(e);
+    throw new Error(`Failed to resolve trust settings. ${message}`, { cause: e });
```
tmathern, line 211: This may be a breaking change for people who did error message parsing (if anyone did)

tmathern (approved):

PR #118, opened 2026-06-04
fix: Merge per-call settings with global settings for Reader and Builder

ale-adobe:


packages/c2pa-web/src/lib/builder.ts
```diff
@@ -183,9 +184,10 @@ export interface ManifestAndAssetBytes {
 
 /**
  * @param worker - Worker (via WorkerManager) to be associated with this reader factory.
+ * @param globalSettings - Optional global settings to be used for all builders.
  * @returns A {@link BuilderFactory} object containing builder creation methods.
  */
-export function createBuilderFactory(worker: WorkerManager): BuilderFactory {
+export function createBuilderFactory(worker: WorkerManager, globalSettings?: Settings): BuilderFactory {
```
tmathern:
```suggestion
export function createBuilderFactory(worker: WorkerManager, settings?: Settings): BuilderFactory {
```
To prepare people to the fact they won't be global forever! (Upcoming changes...)

packages/c2pa-web/src/lib/builder.ts
```diff
@@ -16,6 +16,7 @@ import type {
   Ingredient,
   ManifestDefinition
 } from '@contentauth/c2pa-types';
+import { merge } from 'ts-deepmerge';
```
tmathern: Note: Not exposed here, but Rust base could handle settings merge!

packages/c2pa-web/src/lib/builder.ts
```diff
@@ -194,7 +196,8 @@ export function createBuilderFactory(worker: WorkerManager): BuilderFactory {
 
   return {
     async new(settings?: Settings) {
-      const settingsJson = settings && (await settingsToWasmJson(settings));
+      const effectiveSettings = settings ? merge(globalSettings ?? {}, settings) : globalSettings;
```
tmathern: Make those 2 lines a function and refactor?

tmathern (commented):
Rename `globalSettings` -> `settings` to start managing expectation that those won't soon be "global" (as in thread-local style) anymore.

scouten-adobe (approved):
Reasonable short-term solution; as noted, we'll want to more fully embrace the `Context` API from the underlying Rust SDK.

tmathern (approved):

PR #119, opened 2026-06-04
feat(c2pa-web): add workerSrc config option for CSP-restricted environments

JD771:
## Summary

- Add `workerSrc?: string` to `Config` interface in `createC2pa` — when provided, the worker is loaded from this HTTPS URL instead of an inline blob URL
- Thread `workerSrc` through `createWorkerManager`; fall back to `InlineWorker` (existing blob behaviour) when omitted — fully backwards compatible
- Add `worker: 'src/lib/worker.ts'` as a vite build entry so the worker script is emitted as `dist/worker.js` for consumers to host
- Expose `./worker` package export so consumers can resolve the path via `@contentauth/c2pa-web/worker`

## Motivation

Environments with a strict Content Security Policy (`worker-src 'self'`) block blob: worker URLs. This change allows consumers to host the worker script at an HTTPS URL on the same origin as their app and pass it via `workerSrc`, bypassing the CSP restriction while keeping the existing inline blob as the default for unconstrained environments.

## Usage

```ts
const c2pa = await createC2pa({
  wasmSrc: 'https://cdn.example.com/c2pa_bg.wasm',
  workerSrc: 'https://cdn.example.com/c2pa_worker.js', // hosted from @contentauth/c2pa-web/worker
});
```

## Test plan

- [ ] Existing tests pass (no behaviour change when `workerSrc` is omitted)
- [ ] Manual test: pass a valid `workerSrc` URL and verify worker initialises without blob URL
- [ ] Manual test: omit `workerSrc` and verify existing blob fallback still works

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/c2pa-web/src/lib/worker/workerManager.ts
```diff
@@ -34,10 +35,12 @@ export interface CreateWorkerManagerConfig {
 export async function createWorkerManager(
   config: CreateWorkerManagerConfig
 ): Promise<WorkerManager> {
-  const { wasm, settingsString } = config;
+  const { wasm, settingsString, workerSrc } = config;
   let signerRequestId = 0;
 
-  const worker = new Worker();
+  const worker = workerSrc
+    ? new Worker(workerSrc, { type: 'module' })
```
tmathern: Suggestion: Do some `workerSrc` string validation here. At the very least, you'd expect a HTTPS. We may also want (but this is more complex) to add validations on a domain by passing in a validation function.

packages/c2pa-web/src/lib/c2pa.ts
```diff
@@ -18,6 +18,16 @@ export interface Config {
    */
   wasmSrc: string | WebAssembly.Module;
 
+  /**
+   * URL to the worker script. When provided the worker is loaded from this
+   * HTTPS URL instead of an inline blob URL, which is required in environments
```
tmathern: That rule that it is a "HTTPS URL" is just a comment now, you may want to enforce that a little bit more strictly

packages/c2pa-web/src/lib/c2pa.ts
```diff
@@ -18,6 +18,16 @@ export interface Config {
...
+   * with a strict Content Security Policy that disallows blob: worker sources.
+   *
+   * Host the file exported as `@contentauth/c2pa-web/worker` alongside your
+   * application and pass its URL here.
+   */
+  workerSrc?: string;
```
tmathern: How do we make sure the versions do not diverge if you self-host?

tmathern (changes requested):

tmathern:
@JD771 Please also make sure to sign the Adobe CLA.

packages/c2pa-web/src/lib/c2pa.ts
klynsaha: @tmathern this is the same worker exposed by this package, so its versioning and updates are managed through the package version.

packages/c2pa-web/src/lib/worker/workerManager.ts
ale-adobe: Agreed, let's add some basic URL validation on the provided path before loading it into a worker.

We can leverage the `URL`  constructor to do basic sanity checks to make sure it's structurally a valid URL.

I'm thinking something like:

```
let url: URL;
try {
  url = new URL(workerSrc);
} catch {
  throw new Error(`Invalid worker source URL provided: ${workerSrc}`);
}

if (url.protocol !== "https:") {
  throw new Error(`Worker source URL must use https, but got ${url.protocol}`);
}

const workerSrcUrl = url.toString();
const worker = new Worker(workerSrc, { type: 'module' });
```

ale-adobe (changes requested):

JD771:
Closing to reopen as a fresh PR.

PR #123, opened 2026-06-05
feat(c2pa-web): add workerSrc config option for CSP-restricted environments

JD771:
## Summary

- Add `workerSrc?: URL` to the `Config` interface in `createC2pa` — when provided, the worker is loaded from this HTTPS URL instead of an inline blob URL
- Thread `workerSrc` through `createWorkerManager`; fall back to `InlineWorker` (existing blob behaviour) when omitted — fully backwards compatible
- Add `validateWorkerSrc` which enforces https protocol before the URL is passed to the Worker constructor, preventing insecure worker sources
- Worker is created as a classic script (no `type: 'module'`) since the hosted file is a self-contained IIFE

## Motivation

Environments with a strict Content Security Policy (`worker-src 'self'`) block blob: worker URLs. This change allows consumers to host the worker script at an HTTPS URL on the same origin as their app and pass it via `workerSrc`, bypassing the CSP restriction while keeping the existing inline blob as the default for unconstrained environments.

## Usage

```ts
const c2pa = await createC2pa({
  wasmSrc: 'https://cdn.example.com/c2pa_bg.wasm',
  workerSrc: new URL('https://cdn.example.com/c2pa_worker.js'),
});
```

## Test plan

- [ ] Existing tests pass (no behaviour change when `workerSrc` is omitted)
- [ ] Manual test: pass a valid `workerSrc` URL and verify worker initialises without blob URL
- [ ] Manual test: omit `workerSrc` and verify existing blob fallback still works
- [ ] Manual test: pass a non-https URL and verify `validateWorkerSrc` throws

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tmathern (approved):

ale-adobe (approved):

tmathern:
@JD771 This seems to be stuck on the jobs. I'd suspect it has to do with the Claude commits.
Since you are an Adobe employee also, I opened https://github.com/contentauth/c2pa-js/pull/126/ with the changes.

PR #128, opened 2026-06-09
feat: bundle c2pa_worker.js as self-contained ESM for direct workerSrc use

klynsaha:
## Summary

- Add a `closeBundle` hook that re-bundles the worker using Vite's `build()` API, inlining `highgain` and all internal imports so consumers can load it directly via `workerSrc` without needing their own bundler to resolve bare specifiers
- Rename worker build entry to `c2pa_worker` so the output is `dist/c2pa_worker.js`
- Pass `{ type: 'module' }` when constructing a `Worker` from `workerSrc`, required for ESM module workers

### DIFF
--- a/packages/c2pa-web/package.json
+++ b/packages/c2pa-web/package.json
@@ -27,10 +27,10 @@
       "import": "./dist/inline.js",
       "default": "./dist/inline.js"
     },
-    "./worker": {
+    "./c2pa_worker": {
       "development": "./src/lib/worker.ts",
-      "import": "./dist/worker.js",
-      "default": "./dist/worker.js"
+      "import": "./dist/c2pa_worker.js",
+      "default": "./dist/c2pa_worker.js"
     }
   },
   "publishConfig": {
@@ -47,9 +47,9 @@
         "import": "./dist/inline.js",
         "default": "./dist/inline.js"
       },
-      "./worker": {
-        "import": "./dist/worker.js",
-        "default": "./dist/worker.js"
+      "./c2pa_worker": {
+        "import": "./dist/c2pa_worker.js",
+        "default": "./dist/c2pa_worker.js"
       }
     }
   },
--- a/packages/c2pa-web/vite.config.ts
+++ b/packages/c2pa-web/vite.config.ts
@@ -6,7 +6,7 @@
  * accordance with the terms of the Adobe license agreement accompanying
  * it.
  */
-import { defineConfig, Plugin } from 'vite';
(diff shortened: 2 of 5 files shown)


klynsaha:
@tmathern Could you please help with the review

packages/c2pa-web/package.json
```diff
@@ -27,10 +27,10 @@
       "import": "./dist/inline.js",
       "default": "./dist/inline.js"
     },
-    "./worker": {
+    "./c2pa_worker": {
```
tmathern, line 30: Why the name changes?

klynsaha: We already have a WASM file named `c2pa_bg.wasm`, so having a similarly specific name for the worker seems reasonable. It would also avoid requiring clients to rename the file when hosting it, since `worker.js` is a very generic name and not specific to C2PA, thoughts?

tmathern: This is a breaking change compared to previous release. Please add it as note to the changeset. I think it is currently okayish since you suggested previous change too, but still a noteworthy breaking change.

packages/c2pa-web/vite.config.ts
```diff
@@ -103,6 +102,23 @@ function createBuildPlugin(): Plugin {
 
       mkdirSync(pkgResourceDir, { recursive: true });
       linkSync(wasmResourcePath, pkgResourcePath);
+    },
+    async closeBundle() {
+      await build({
```
tmathern, line 107: Fine for explicit build, but that does multiple build runs

klynsaha: yeah, the worker build is separate because the primary build step compiles all entries under a single shared `rollupOptions.external` list - `['@contentauth/c2pa-types', 'highgain', 'ts-deepmerge']`. There's no per-entry way to say "externalize these deps for index and inline, but bundle them into c2pa_worker."

`closeBundle` lets us run a completely separate Rollup/Vite invocation with its own `rollupOptions`, so the worker gets its own externals config (`['@contentauth/c2pa-types']` only, leaving `highgain` and `ts-deepmerge` to be inlined).

The alternative would be to split the worker into a separate `vite.config.worker.ts` and run two build commands in CI, but the `closeBundle` hook achieves the same thing without changing the build pipeline. 

Do you want me to update this to use a separate worker config?

tmathern: No, it's fine. Just noting

tmathern (approved):

PR #143, opened 2026-06-23
fix: normalize object settings so trust/verify config isn't silently dropped

danspicytaco:
## Problem

`Reader.fromAsset`, `Reader.fromManifestDataAndAsset`, and three `Builder` methods (`new`, `withJson`, `fromArchive`) serialize object-form `settings` with `JSON.stringify`. That leaves camelCase keys (e.g. `verifyTrustList`, `trustAnchors`) in place — keys that c2pa-rs does **not** recognize and silently drops.

The helper functions `createTrustSettings` / `createVerifySettings` / `mergeSettings` all produce camelCase objects, and their docstrings say the result "can be passed to `Reader`/`Builder`". In practice the settings have no effect unless the caller manually pre-converts them with `settingsToJson`. The failure is silent — e.g. trust verification appears configured but is quietly ignored.

## Fix

Add a shared `normalizeSettings(settings)` helper in `Settings.ts` and use it at all five call sites:

- A **string** is passed through unchanged.
- An **object** goes through `settingsToJson` (camelCase → snake_case), the same conversion callers would otherwise have to do by hand.

Backward-compatible: already-snake_case keys are idempotent under the converter, and string callers are unaffected. Adds regression tests to `Settings.spec.ts` asserting that camelCase keys are converted (and do **not** survive) and that `normalizeSettings` matches `settingsToJson` for object input.

## Validation

The change is a verbatim re-application of a patch previously validated against `c2pa-node-v2` @ 0.5.5 (`tsc -b` clean, `eslint` clean, Settings spec 21/21 passing), onto the now-migrated `packages/c2pa-node` copy of the same source. Runtime behavior was confirmed empirically: a camelCase settings object produced an *untrusted* result, while the equivalent `settingsToJson` string produced a *trusted* result — this PR makes the object form behave like the string form.

### DIFF
--- a/packages/c2pa-node/js-src/Settings.spec.ts
+++ b/packages/c2pa-node/js-src/Settings.spec.ts
@@ -10,6 +10,7 @@ import {
   createVerifySettings,
   mergeSettings,
   settingsToJson,
+  normalizeSettings,
   loadSettingsFromFile,
   loadSettingsFromUrl,
 } from "./Settings.js";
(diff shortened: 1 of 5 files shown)


danspicytaco:
hey @ale-adobe 👋 could I get a review of this PR please? It is my first PR to this repo so I am unsure about the process ~ please let me know if there I am missing anything. Thank you!!

packages/c2pa-node/js-src/Settings.spec.ts
```diff
@@ -322,4 +323,42 @@ verify_after_sign = false`;
...
+    it("returns undefined when no settings are provided", () => {
+      expect(normalizeSettings(undefined)).toBeUndefined();
+    });
+
+    it("passes a string through unchanged", () => {
+      const json = '{"verify":{"verify_trust":true}}';
+      expect(normalizeSettings(json)).toBe(json);
+    });
+
+    it("converts a camelCase settings object to snake_case JSON", () => {
```
tmathern, line 337: Can you also add a test verifying this claim from the PR description: `already-snake_case keys are idempotent under the converter, and string callers are unaffected.`?

ale-adobe:
Hey @DanSpicyTaco,

First off, thanks for identifying this issue, and I appreciate you taking the time to put together a PR!

This is definitely a gap in the Node SDK. Fortunately, it isn't an issue in the Web SDK, since that already performs the necessary normalization.

That said, my longer-term goal is for both SDKs in this repository to share a common set of utility functions for lower-level functionality such as settings normalization, rather than maintaining effectively the same logic in two separate implementations. That's a somewhat larger refactor than the changes in this PR, and it's something I already have planned to tackle in the near future.

With that in mind, I'm going to hold off on merging this PR. I want to avoid introducing code that will likely be replaced shortly as part of that shared implementation.

Thanks again for taking the time to investigate this and contribute a fix. It helped confirm that closing the remaining gaps between the two SDKs is an important priority.

danspicytaco:
Thanks @ale-adobe, that makes sense. I’ll close this for now, rather than keeping a Node-only fix open if the shared settings utility work what we want here. Glad it helped confirm the gap! Happy to help rework this or help out in other areas too if that's useful 😄

PR #145, opened 2026-06-26
fix: sanitize manifestStore() output against __proto__ manifest labels

ale-adobe:
`serde_wasm_bindgen` serializes Rust `HashMap` entries using bracket assignment (`obj[key] = value`). For the key `"__proto__"`, JavaScript intercepts this as `[[SetPrototypeOf]]` rather than creating an own property, so the manifest ends up in the prototype chain of the manifests object. When the worker returns the result over `postMessage`, structured clone only copies own properties, causing the manifest to silently disappear.

This PR adds a helper function `sanitizeManifestStore` which is used in the worker's `reader_manifestStore` handler. It rebuilds manifests as a null-prototype object (safe for arbitrary keys) and recovers any manifest displaced into the prototype chain before the value is structured-cloned.

Includes unit tests that directly construct the poisoned object state `serde_wasm_bindgen` would produce and verify the repair. Note: `c2pa-rs` itself rejects `__proto__` as a label through the normal builder API, so the fix is defense-in-depth against pre-existing signed assets from other tools or older library versions.

### DIFF
--- a/packages/c2pa-web/src/lib/worker/sanitizeManifestStore.ts
+++ b/packages/c2pa-web/src/lib/worker/sanitizeManifestStore.ts
@@ -1,0 +1,37 @@
+/**
+ * Copyright 2026 Adobe
+ * All Rights Reserved.
+ *
+ * NOTICE: Adobe permits you to use, modify, and distribute this file in
+ * accordance with the terms of the Adobe license agreement accompanying
+ * it.
+ */
+
+/**
+ * Rebuilds the manifest store's manifests map as a null-prototype object so
+ * that reserved property names like "__proto__" are safe to use as keys.
+ *
+ * serde_wasm_bindgen uses bracket assignment when serializing HashMap entries.
+ * For the key "__proto__", `obj["__proto__"] = value` triggers [[SetPrototypeOf]]
+ * instead of creating an own property, putting that manifest in the prototype
+ * chain rather than the object itself. postMessage's structured clone only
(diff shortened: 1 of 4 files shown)


tmathern:
What about other potential properties like `constructor`. Do they also need special handling then?

packages/c2pa-web/src/lib/worker/sanitizeManifestStore.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * serde_wasm_bindgen uses bracket assignment when serializing HashMap entries.
+ * For the key "__proto__", `obj["__proto__"] = value` triggers [[SetPrototypeOf]]
+ * instead of creating an own property, putting that manifest in the prototype
+ * chain rather than the object itself. postMessage's structured clone only
+ * copies own properties, so it would disappear in transit. We detect this by
+ * checking whether the prototype was changed, and restore the manifest as an
+ * own property before the value leaves the worker.
+ */
+export function sanitizeManifestStore(store: any): any {
```
tmathern, line 22: I have a suspicion this function will become more complex over time to handle more cases like this!

tmathern (approved):

tmathern:
(Don't forget the changeset)

ale-adobe:
> What about other potential properties like constructor. Do they also need special handling then?

No, `__proto__` is a special case.

Issue #147, opened 2026-07-06
Byte-based verification entry point for non-browser runtimes (Node, Cloudflare Workers, Deno)


SilyNoMeta:
## Problem

The only verification entry points on `WasmReader` are `fromBlob` / 
`fromBlobFragment`, which read input through `FileReaderSync` — a 
WorkerGlobalScope-only API. `createC2pa` in c2pa-web additionally 
requires the `Worker` constructor. Neither exists in server-side / 
edge runtimes (Node, Cloudflare Workers workerd, Deno, Bun), which 
makes C2PA verification currently impossible outside a browser, even 
though the underlying wasm (with `rust_native_crypto`) has no browser 
dependency for the crypto itself.

Context: we run pay-per-call utility APIs for AI agents on Cloudflare 
Workers (402utils.com) and want to offer Content Credentials 
verification — especially relevant with the EU AI Act Art. 50 
transparency obligations applying from 2026-08-02.

## Proposal

Add a `fromBytes(format: string, bytes: Uint8Array, settings?)` 
sibling to `fromBlob`, feeding the existing private 
`WasmReader::from_stream` helper a `Cursor<Vec<u8>>` instead of a 
`BlobStream`. ~15 lines, no new dependency, no browser API, identical 
return type and validation behavior.

## Evidence

We have this working: verified 5 cases (valid manifest, no manifest, 
corrupted claim, corrupted signature, corrupted asset bytes) with 
byte-identical results between Node and real workerd (`wrangler dev`). 
All tampering is correctly detected; `getRandomValues` resolves via 
`globalThis.crypto`; no hidden browser-API layer appeared.

A PR is ready on my fork (`feat/wasm-reader-from-bytes`: method + two 
`wasm_bindgen_test`s + signed fixture + changeset) — happy to open it 
if this direction works for you. CLA signed.

Note: this touches the same files as #137 (BlobStream perf); I'll 
rebase as needed.

PR #137, opened 2026-06-17
fix: eliminate repeated JS/WASM boundary crossings in BlobStream

saeedrahimi:
Fixes #93

PNG files with lots of chunks (AI-generated images from Microsoft Designer or Nano Banana 2, for example) were painfully slow to read. The original `BlobStream` called `Blob.slice()` + `FileReaderSync.readAsArrayBuffer()` on every single `read()`, crossing the JS/WASM boundary hundreds of times per file.

This introduces a dual-strategy `BlobStream`:

- Files under 50 MB are read into memory upfront in one shot, so subsequent `read()` calls just slice an `ArrayBuffer` with no JS/WASM crossing.
- Files 50 MB and over fall back to the original lazy per-read slice strategy to avoid doubling memory usage.

For typical files, JS/WASM crossings drop from O(N chunks) to O(1). PNG read times go from 8+ seconds to under a second for affected files.

Issue #93, opened 2026-03-17
reader.fromBlob is very slow on some PNG files (3+ seconds)


cesrahimi:
**Update:**  Added Nano Banana 2 image and updated the timings

## Description

Reading C2PA manifests from some PNG files using `reader.fromBlob` is significantly slower than other formats (e.g., JPEG). Processing a single AI-generated PNG (from Nano Banana 2 and Microsoft Designer) takes **8+ seconds** in the browser, which is not acceptable for interactive use cases.

## Steps to Reproduce

1. Use `createC2pa()` to initialize the SDK
2. Obtain a C2PA-signed PNG file (e.g., generated by Microsoft Designer with AI content credentials)
3. Call `reader.fromBlob("image/png", file)` followed by `reader.manifestStore()`
4. Observe that the total read time exceeds 3 seconds

## Minimal Code

```ts
import { createC2pa } from "@contentauth/c2pa-web";

const sdk = await createC2pa({
  wasmSrc: new URL("@contentauth/c2pa-web/resources/c2pa.wasm", import.meta.url).href,
});

const reader = await sdk.reader.fromBlob(file.type, file); // slow for AI-Generated PNG files
const manifestStore = await reader.manifestStore();
await reader.free();
```

## Expected Behavior

Manifest reading for a PNG file should complete in a comparable timeframe to JPEG files (under 1 second for typical file sizes).

## Actual Behavior

PNG files take 3+ seconds to process. JPEG files of similar size are noticeably faster.

## Environment

- `@contentauth/c2pa-web`: 0.6.1
- Browser: Microsoft Edge Version 145.0.3800.97 (Official build) (64-bit)
- OS: Windows 11
- PNG sources: 
   - Nano Banana 2 (AI-generated image)
   - Microsoft Designer (AI-generated image)

## Additional Context

The same PNG files are parsed by **c2patool CLI v0.26.36** in under **300ms**, making the web SDK roughly **25x slower** for this case. This suggests the bottleneck is specific to the WASM/web implementation rather than inherent to PNG manifest parsing.

The slowness is not limited to reading. Using this PNG as an **ingredient for manifest writing/signing** is also dramatically slower compared to other image types.

**Sample files:** 
- [NanoBanana2.png](https://github.com/user-attachments/assets/e7a6a5d7-0325-4317-b6bb-9c27159c9bc4)  
[...]

PR #148, opened 2026-07-06
feat(c2pa-wasm): add fromBytes verification entry point for non-browser runtimes

SilyNoMeta:
# PR: Add `WasmReader.fromBytes` for non-browser C2PA verification - Refs #147

## Problem

`WasmReader` today exposes only `fromBlob` / `fromBlobFragment`. Both wrap the asset in a
`BlobStream` (`packages/c2pa-wasm/src/stream/blob_stream.rs`) which reads bytes through
`web_sys::FileReaderSync` — a `WorkerGlobalScope`-only Web API. As a result, verification is only
usable inside a browser Web Worker. It cannot run in server/edge JavaScript runtimes that have no
`Blob`/`FileReaderSync`/`Worker`:

- **Cloudflare Workers (workerd)** — no `Worker` constructor, no `FileReaderSync`.
- **Node.js** (outside a DOM), **Deno**, **Bun** — same.

These runtimes already have the asset bytes in memory (an upload, a `fetch` response, a KV/R2
object), so requiring a `Blob` + `FileReaderSync` round-trip is both a portability blocker and,
per #93, a per-read JS↔WASM boundary cost.

## Solution

Add a sibling entry point that takes the bytes directly:

```rust
#[wasm_bindgen(js_name = fromBytes)]
pub async fn from_bytes(
    format: &str,
    bytes: Vec<u8>,
    context_json: Option<String>,
) -> Result<WasmReader, JsString> {
    let stream = Cursor::new(bytes);
    WasmReader::from_stream(format, stream, context_json).await
}
```

It reuses the **existing** private `from_stream(format, impl Read + Seek + Send, context_json)`
helper (the same path `fromBlob` uses), feeding it a `std::io::Cursor<Vec<u8>>` instead of a
`BlobStream`. No new dependency, no browser API, and the returned `WasmReader` and its
validation results are identical to `fromBlob`. `Cursor`, `Read`, and `Seek` are already imported.

```ts
// TypeScript
const reader = await WasmReader.fromBytes("image/jpeg", new Uint8Array(bytes));
```

## Runtimes unblocked

Node.js, Deno, Bun, and Cloudflare Workers (workerd). Verified end-to-end on **workerd** (via
`wrangler dev`) and **Node.js** with the pure-Rust crypto build (`rust_native_crypto`): a valid
asset validates (signature + hash bindings), an asset with no C2PA errors with `JumbfNotFound`, a
tampered signature yields `claimSignature.mismatch` (Invalid), and a tampered asset yields
`assertion.dataHash.mismatch` (Invalid) — byte-identical between Node and workerd. `getRandomValues`
resolves via `globalThis.crypto` on workerd; no `FileReaderSync`/`Blob`/`Worker` is touched.

## Tests

Adds two `wasm_bindgen_test`s in `wasm_reader.rs` that run in the **default (non-worker)** context
(unlike the `BlobStream` tests, which require `run_in_dedicated_worker`) — this is itself the proof
that `fromBytes` needs no Web Worker:

- `from_bytes_reads_active_manifest` — a signed JPEG (`tests/fixtures/C.jpg`) yields an active manifest.
- `from_bytes_without_manifest_errors` — a buffer with no C2PA data is rejected, not silently accepted.

## Notes

- **Overlaps with #137** (BlobStream perf) — both touch `wasm_reader.rs` / `blob_stream.rs`, so a
  rebase may be needed depending on merge order. The changes are complementary: #137 optimizes the
  Blob path; this adds a Blob-less path. (#137 already reads the whole Blob into a `Cursor` for
  ≤50 MB assets, so the `bytes → Cursor` primitive is a natural fit.)
- **c2pa-web wiring is intentionally out of scope.** `c2pa-web` runs the reader inside a Web Worker;
  the value of `fromBytes` is for consumers that bypass that Worker. Exposing it through the
  `c2pa-web` Worker RPC (`reader_fromBytes`) can be a follow-up.
- `WasmReader::from_stream` currently uses the deprecated `Reader::from_stream_async`; this PR reuses
  that existing helper rather than introducing new deprecation. Migrating the helper to
  `Reader::from_context(ctx).with_stream_async` can be a separate cleanup.

## Checklist (contributor)

- [x] Individual Adobe CLA signed (one-time)
- [x] Changeset added (`.changeset/add-wasm-reader-frombytes.md`, minor bump for `@contentauth/c2pa-wasm`)
- [x] Adobe license header present on touched files
- [ ] `nx test c2pa-wasm` green in CI (headless Chromium)


tmathern:
For node-based workers, why not use https://github.com/contentauth/c2pa-js/tree/main/packages/c2pa-node ?

SilyNoMeta:
Good call, i actually dug into that one before going the wasm route! 
c2pa-node builds c2pa-rs into a native c2pa.node addon (via cargo + a postinstall), plus it pulls in sharp/libvips, so it needs a full Node runtime with native addon loading. Workers (workerd) only runs JS + WASM, no native .node stuff, so it can't load there (same story for Deno/Bun by default)

ale-adobe:
Hi @SilyNoMeta, thanks for your PR! Would you mind updating to latest `main` for us?

Separate question: is there a particular use case where a Node user would want to use this rather than using `c2pa-node` directly? As it currently stands, I'm seeing this as mostly a benefit to the other use cases that you've pointed out, such as Deno and Bun.

SilyNoMeta:
Hi @ale-adobe and thanks ! 
I've updated the branch with the latest main; all 18 WASM tests pass locally in headless Chrome.
My actual deployment target is Cloudflare Workers (workerd), where c2pa-node cannot run because it requires a native .node addon. That's precisely why I went with c2pa-wasm, and added fromBytes to remove the browser-only I/O requirement.

I agree that for a Node-only application, c2pa-node is the natural choice. Node was useful during my prototype to test the same WASM build used on workerd, but the primary motivation for this PR is enabling verification directly in Cloudflare Workers.

.changeset/add-wasm-reader-frombytes.md
```diff
@@ -0,0 +1,5 @@
+---
+"@contentauth/c2pa-wasm": minor
```
ale-adobe: Let's make this a patch update. The changes are additive and non-breaking. Since we are still pre-1.0, we are reserving minor version updates for breaking changes at this time.

ale-adobe (changes requested):
Thanks for updating! Small request to update the changeset.

.changeset/add-wasm-reader-frombytes.md
SilyNoMeta: Done!

ale-adobe (approved):
Thanks for your contribution!

PR #153, opened 2026-07-15
[POC] Migrate from Neon to koffi

ale-adobe:


Cargo.toml
```diff
@@ -4,7 +4,6 @@ resolver = '2'
 members = [
 	'packages/c2pa-wasm',
 	'packages/c2pa-types',
-	'packages/c2pa-node',
```
ale-adobe, line 7: `c2pa-node` would no longer be a Rust crate, just like `c2pa-web`.

packages/c2pa-node/js-src/native/context.ts
```diff
@@ -0,0 +1,70 @@
...
+// each license.
+
+import { getLib } from "./lib.js";
+import { checkInt, checkPtr } from "./error.js";
+import type { C2paSettings } from "../types.d.ts";
+
+/**
+ * A native C2paContext*, required to construct a Reader or Builder.
+ * This package's public API takes settings per-call (as a JSON/TOML string
+ * or object) rather than exposing the context concept directly, so this
```
tmathern, line 21: If this moves closer to the C FFI, maybe consider exposing the `Context` too? (Potentially as overloads?). Contexts are reusable, and can hold settings and signers.
(In any case it probably would be good to make Context become an exposed class/object).

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -11,50 +11,122 @@
 // specific language governing permissions and limitations under
 // each license.
 
-import type {
-  BuilderIntent,
-  C2paReason,
-  Ingredient,
-  Manifest,
-} from "@contentauth/c2pa-types";
+import type { BuilderIntent, C2paReason, Ingredient, Manifest } from "@contentauth/c2pa-types";
```
tmathern, line 14: Builder object pointer need also a c2pa_free somewhere (dispose?). I don't see it here, but maybe it is done generally somewhere else?

packages/c2pa-node/js-src/Reader.ts
```diff
@@ -13,54 +13,127 @@
 
 import type { Manifest, ManifestStore } from "@contentauth/c2pa-types";
 
-import { getNeonBinary } from "./binary.js";
+import { getLib, decodeAndFree, callAsync } from "./native/lib.js";
```
tmathern, line 16: Reader object pointer needs also a c2pa_free somewhere (dispose?). I don't see it here, but maybe it is done generally somewhere else?

packages/c2pa-node/js-src/Reader.ts
```diff
@@ -13,54 +13,127 @@
...
+        "Failed to create C2paReader",
+      );
+      const stream = C2paStream.fromSource(asset);
+      try {
+        const newPtr = checkPtr(
+          getLib().c2pa_reader_with_stream(readerPtr, mimeTypeOf(asset) ?? "", stream.ptr),
+          "Failed to read asset",
+        );
+        const reader = new Reader(newPtr);
+        reader._ctx = ctx;
```
tmathern, line 91: The Reader should probably release this reference when being cleaned up/released/disposed. (Is there a dispose() for the Reader?)

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -188,83 +280,102 @@ export class Builder implements BuilderInterface {
...
+          mimeTypeOf(input) ?? "",
+          src.ptr,
+          dest.ptr,
+          nativeSigner.ptr,
+          manifestBytesOut,
+        ),
+        "Failed to sign",
+      );
+      C2paStream.finalizeDestination(output, dest, size);
+      return decodeBytesAndFree(manifestBytesOut[0], size);
```
tmathern, line 319: Since the bytes were allocated, you need to make sure they would get freed even if this throws anywhere in the path towards this casee (otherwise the memory leaks size of bytes)

packages/c2pa-node/js-src/native/stream.ts
```diff
@@ -0,0 +1,352 @@
...
+          // Fast path: backends that already hold their bytes in JS memory
+          // (BufferBackend) can hand back a view directly, skipping the
+          // scratch-buffer copy below.
+          const view = this._backend.readView?.(n);
+          if (view !== undefined) {
+            if (view.length === 0) return 0;
+            koffi.encode(outPtr, "uint8_t", view, view.length);
+            return view.length;
+          }
+          if (this._scratch.length < n) this._scratch = Buffer.alloc(n);
```
tmathern, line 227: Does this growth needs to be capped, potentially?

packages/c2pa-node/js-src/native/signer.ts
```diff
@@ -0,0 +1,157 @@
...
+      ): number => {
+        try {
+          const n = toNum(len);
+          const maxOut = toNum(outLen);
+          if (n <= 0 || n > MAX_DATA || maxOut <= 0) return -1;
+
+          const arr = koffi.decode(dataPtr, "uint8_t", n) as number[];
+          const signature = sign(Buffer.from(arr));
+
+          const actual = Math.min(signature.length, maxOut);
```
tmathern, line 91: This can silently truncate length. Shouldn't it fail instead?

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -64,122 +136,142 @@ export class Builder implements BuilderInterface {
...
-        this.builder,
-        ingredientJson,
-        ingredient,
-      );
-    } else {
-      return getNeonBinary().builderAddIngredient.call(
-        this.builder,
-        ingredientJson,
+    if (!ingredient) {
+      throw notImplemented("addIngredient (without a source asset)");
```
tmathern, line 205: What are you trying to get implemented here? Adding an ingredient archive? What is an ingredient without an asset?
(Note: depending on the answer, it may be a matter of wrangling JSON and resources).

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -188,83 +280,102 @@ export class Builder implements BuilderInterface {
...
-      value,
-    );
+  updateManifestProperty(_property: string, _value: ClaimVersion): void {
+    throw notImplemented("updateManifestProperty");
   }
 
-  addRedaction(uri: string, reason: C2paReason): void {
-    getNeonBinary().builderAddRedaction.call(this.builder, uri, reason);
+  addRedaction(_uri: string, _reason: C2paReason): void {
+    throw notImplemented("addRedaction");
```
tmathern, line 375: This is a matter of building the right JSON, or using future APIs from https://github.com/contentauth/c2pa-rs/pull/2281 exposed themselves through C FFI.
Redactions are URIs set on the Builder JSON on a redactions field (for now). The linked PR may introduce other APIs.

packages/c2pa-node/js-src/native/context.ts
```diff
@@ -0,0 +1,70 @@
...
+    this.ptr = ptr;
+  }
+
+  static create(settings?: C2paSettings): Context {
+    if (settings === undefined) {
+      return new Context(checkPtr(getLib().c2pa_context_new(), "Failed to create C2paContext"));
+    }
+
+    const settingsStr =
+      typeof settings === "string" ? settings : JSON.stringify(settings);
```
tmathern, line 38: In theory, JSON.stringify can throw if the settings object is circular. I would protect against that here and rethrow an error accordingly.

packages/c2pa-node/js-src/native/context.ts
```diff
@@ -0,0 +1,70 @@
...
+    const format = settingsStr.trim().startsWith("{") ? "json" : "toml";
```
tmathern, line 39: I would advise against supporting toml here. Consider asking only for JSON settings here, since json is quite natural in javascript. Except if Node already uses and prefers toml too (that would be the exception, but maybe an overload to provide in node only then).

packages/c2pa-node/js-src/native/stream.ts
```diff
@@ -0,0 +1,352 @@
...
+    const end = this._pos + data.length;
+    this._ensureCapacity(end);
+    data.copy(this._buf, this._pos);
+    this._pos += data.length;
+    if (this._pos > this._size) this._size = this._pos;
+    return data.length;
+  }
+
+  seek(offset: number, mode: number): number {
+    const newPos = seekTo(offset, mode, this._pos, this._size);
```
tmathern, line 121: I would add checks that position is not outside of size.
(This is useful because sometimes the streams rewind, sometimes not. THis is to make sure writes are inside bounds!)

packages/c2pa-node/js-src/Signer.ts
```diff
@@ -11,104 +11,191 @@
...
+   * Build the native signer used by Builder.signAsync(). koffi's signer
+   * callback runs synchronously on the main thread and cannot await real
+   * async work (e.g. a network round-trip). This only works if the
+   * callback's promise happens to already be settled by the time koffi
+   * calls it synchronously, which in practice means it will fail for any
+   * callback that performs genuine async I/O.
+   */
+  nativeSigner(): NativeSigner {
+    if (!this._native) {
+      // `this._callback` is typed `(data: Buffer) => Promise<Buffer>`, so it
```
tmathern, line 157: Does this mean a sync API is needed?

packages/c2pa-node/js-src/Signer.ts
```diff
@@ -11,104 +11,191 @@
...
+      // completed. koffi's signer callback must return synchronously, so a
+      // genuine async JS callback signer cannot be bridged through it at
+      // all. This is not a bug to fix here — it's a fundamental mismatch
+      // between koffi's synchronous callback model and a Promise-based
+      // signer API. native/signer.ts's koffi callback wrapper swallows
+      // thrown errors (returns -1) since koffi callbacks can't propagate JS
+      // exceptions, so this message is stashed on `lastSyncError()` for
+      // Builder to surface instead of a generic "COSE signature invalid"
+      // error from the native signing failure.
+      const syncBridge = (): Buffer => {
```
tmathern, line 169: Signers can be reused for multiple calls (in theory) and multiple Builders. Therefore, you need to make sure `_lastSyncError` gets read at the right time, and also cleaned so the Signer stays usable.

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -188,83 +280,102 @@ export class Builder implements BuilderInterface {
...
   }
 
-  getHandle(): NeonBuilderHandle {
-    return this.builder;
+  getHandle(): unknown {
```
tmathern, line 378: Can this private/hidden? Or at least have better typing?

packages/c2pa-node/js-src/native/context.ts
```diff
@@ -0,0 +1,70 @@
...
+    const settingsPtr = checkPtr(
+      getLib().c2pa_settings_new(),
+      "Failed to create C2paSettings",
+    );
+    checkInt(
+      getLib().c2pa_settings_update_from_string(settingsPtr, settingsStr, format),
+      "Failed to apply settings",
+    );
+
+    const builderPtr = checkPtr(
```
tmathern, line 50:
```suggestion
    const contextBuilderPtr = checkPtr(
```
(so it's not confused in any way with a Builder pointer object by people reading this file!)

packages/c2pa-node/js-src/native/error.ts
```diff
@@ -0,0 +1,85 @@
...
+// never the worker thread that set the error. So async failures can only
+// report a generic message, not the detailed reason the sync API provides.
+
+/** Check integer return value from an async call; throws a generic error
+ * (no detailed message — see note above) if negative. */
+export function checkIntAsync(
+  result: number | bigint,
+  fallback = "Unknown C2PA error",
+): number {
+  const n = typeof result === "bigint" ? Number(result) : result;
```
tmathern, line 72: This is similar to code above in the same file, at least you can refactor.

packages/c2pa-node/js-src/native/error.ts
```diff
@@ -0,0 +1,85 @@
...
+  return n;
+}
+
+/** Check pointer return value from an async call; throws a generic error
+ * (no detailed message — see note above) if null/falsy. */
+export function checkPtrAsync<T>(
+  ptr: T | null | undefined | false | 0,
+  fallback = "Unknown C2PA error",
+): T {
+  if (!ptr) throw new C2paError(fallback);
```
tmathern, line 83: What guarantees that pointer, if a real memory address, is valid in another thread? Can those addresses cross threads? Who owns freeing them? The creating thread or the using one?

packages/c2pa-node/js-src/native/lib.ts
```diff
@@ -0,0 +1,332 @@
...
+
+// ── Opaque types (registered once globally by name) ─────────────────────────
+
+koffi.opaque("C2paStream");
+koffi.opaque("C2paReader");
+koffi.opaque("C2paBuilder");
+koffi.opaque("C2paSigner");
+koffi.opaque("C2paContext");
+koffi.opaque("C2paSettings");
+koffi.opaque("C2paContextBuilder");
```
tmathern, line 36: Move it next to the C2paContext registration

packages/c2pa-node/js-src/native/lib.ts
```diff
@@ -0,0 +1,332 @@
...
+      "c2pa_identity_signer_create",
+      "C2paSigner *",
+      ["C2paSigner *", "C2paSigner *", "char **", "char **"],
+    ),
+  };
+}
+
+// ── Shared helpers ───────────────────────────────────────────────────────────
+
+// MAX_CSTRING_LEN from c2pa.h (1 MB). koffi.decode stops at the null terminator.
```
tmathern, line 292: I am not sure I understand. Is this limit taken into account then, or not?

packages/c2pa-node/js-src/native/signer.ts
```diff
@@ -0,0 +1,157 @@
...
+  /** @internal */ ptr: unknown;
+  private _disposed = false;
+  // Keep the JS callback alive as long as this signer exists.
+  private _cb: IKoffiRegisteredCallback | null = null;
+
+  private constructor(ptr: unknown) {
+    this.ptr = ptr;
+  }
+
+  /** Create a signer from PEM credentials. The private key never leaves the process. */
```
tmathern, line 47: `The private key never leaves the process`. What does this mean? Process as in OS process/thread, or signing process?

packages/c2pa-node/js-src/native/signer.ts
```diff
@@ -0,0 +1,157 @@
...
+   * DER-encoded signature *synchronously* — koffi callbacks run on the main
+   * thread and cannot await real async work (e.g. a network round-trip).
+   */
+  static fromCallback(
+    sign: (data: Buffer) => Buffer,
+    alg: NativeSigningAlg,
+    certs: string,
+    taUrl?: string,
+  ): Signer {
+    const MAX_DATA = 1024 * 1024; // 1 MB safety limit
```
tmathern, line 73: Does this come form c2pa-rs too? Are there limits there to match?

packages/c2pa-node/js-src/native/signer.ts
```diff
@@ -0,0 +1,157 @@
...
+      koffi.pointer(SignerCallbackProto),
+    );
+
+    const ptr = checkPtr(
+      getLib().c2pa_signer_create(null, wrapped, SigningAlgValues[alg], certs, taUrl ?? null),
+      "Failed to create callback signer",
+    );
+
+    const s = new Signer(ptr);
+    s._cb = wrapped;
```
tmathern, line 107: Why not overload the constructor to take a callback, in this case? And handle the potential callback setting in the constructor?

packages/c2pa-node/js-src/native/stream.ts
```diff
@@ -0,0 +1,352 @@
...
+}
+
+/** In-memory backend. Grows on write; the whole asset is held in JS memory. */
+class BufferBackend implements StreamBackend {
+  private _buf: Buffer;
+  private _pos = 0;
+  private _size: number;
+
+  constructor(buf: Buffer, size: number) {
+    this._buf = buf;
```
tmathern, line 91: Consider adding checks for over- and underflow protection for values that are already clearly wrong at this point (fail early).

packages/c2pa-node/js-src/native/stream.ts
```diff
@@ -0,0 +1,352 @@
...
+
+  close(): void {}
+
+  getBytes(): Buffer {
+    return this._buf.subarray(0, this._size);
+  }
+
+  private _ensureCapacity(needed: number): void {
+    if (needed <= this._buf.length) return;
+    const newBuf = Buffer.alloc(Math.max(needed, this._buf.length * 2));
```
tmathern, line 138: I suspect this can fail if the OS can't give you that memory. Might be worth adding some checks and error handling here.

packages/c2pa-node/js-src/native/stream.ts
```diff
@@ -0,0 +1,352 @@
...
+    this._buf.copy(newBuf, 0, 0, this._size);
+    this._buf = newBuf;
+  }
+}
+
+/** File-descriptor backend: reads/writes go straight to disk, no full buffering. */
+class FileBackend implements StreamBackend {
```
tmathern, line 145: Consider adding error handling in case the file can't be read/written etc., which might happen during a run (e.g. file moved or access rights were changed on the fly by something else).

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -11,50 +11,122 @@
...
+ * c2pa-rs/sdk/src/builder.rs): `Builder::add_assertion(label, &impl
+ * Serialize)`, `Builder::add_ingredient(impl Into<Ingredient>)` (no stream
+ * required), and the public `Builder.definition: ManifestDefinition` field
+ * (redactions, property updates, and reading the definition back are all
+ * just reads/writes of that field or `serde_json::to_string(&builder)`,
+ * since `Builder` derives `Serialize`). None of this is Adobe-specific, so
+ * the right place to add C wrappers is upstream in the public,
+ * MIT/Apache-2.0-licensed `c2pa_c_ffi` crate (e.g. following the existing
+ * `c2pa_builder_add_action` as a template) — NOT in Adobe's private
+ * `adobe_api` repo, which would tie this open source package's core
```
tmathern, line 61: (Some info should not be here...).

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -11,50 +11,122 @@
...
+ * currently export these.
+ *
+ * The underlying Rust methods/fields do exist (confirmed against
+ * c2pa-rs/sdk/src/builder.rs): `Builder::add_assertion(label, &impl
+ * Serialize)`, `Builder::add_ingredient(impl Into<Ingredient>)` (no stream
+ * required), and the public `Builder.definition: ManifestDefinition` field
+ * (redactions, property updates, and reading the definition back are all
+ * just reads/writes of that field or `serde_json::to_string(&builder)`,
+ * since `Builder` derives `Serialize`). None of this is Adobe-specific, so
+ * the right place to add C wrappers is upstream in the public,
```
tmathern, line 58: The C FFI currently is small on purpose (types are Builder, Reader, Signer, Context and Stream, which are the main objects).

The question is not about the wrapper, it is about the abstraction to expose to other SDKs and their public APIs: namely, do we want to have an assertion type? What does this type look like? How do we create it? is it a top level type? Or only constructible through a builder pattern? 
Therefore, it would be good to try to determine at least which input params they take (esp. the addAssertion one), so we can see what is needed/how to get to those, and especially if we want a new type, and what it looks like.

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -11,50 +11,122 @@
...
+ * (~/Repos/c2pa-python, same authors, also a ctypes/C-ABI binding over this
+ * exact `c2pa_c_ffi` layer) has the identical gap — its `Builder` class
+ * (src/c2pa/c2pa.py) has no `add_assertion`/`add_redaction`/
+ * `get_manifest_definition`/`update_manifest_property` either, for the same
+ * reason. Every C-ABI language binding is blocked on the same upstream work.
+ * Notably, `c2patool` (the reference CLI, pure Rust, no C ABI) never needed
+ * these either — its whole interface is upfront JSON manifest files, same
+ * pattern c2pa-python's own docs use exclusively.
+ *
+ * This is an open decision, not just a to-do: land the upstream PR (this
```
tmathern, line 74: Not all APIs should be exposed 1:1 like they are in the Rust code base. Some deserve additional abstraction work, some things can be achieved (already) through JSON wrangling (e.g. redactions). The C FFI surface should aime to be flexible, but remain small (aka, the proper abstractions need to be picked, added and used).

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -64,122 +136,142 @@ export class Builder implements BuilderInterface {
...
   }
 
-  addIngredientFromReader(reader: ReaderInterface): Ingredient {
-    const readerHandle = reader.getHandle();
-    const result = getNeonBinary().builderAddIngredientFromReader.call(
-      this.builder,
-      readerHandle,
-    );
-    return JSON.parse(result);
+  addIngredientFromReader(_reader: ReaderInterface): Ingredient {
```
tmathern, line 223: I have an idea why this exists here as API, and this can probably also be done with the other APIs "manually" by using JSON wrangling from the Reader, and adding resources where/when needed.
(Might not need something dedicated from the C FFI, but some more code here).

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -64,122 +136,142 @@ export class Builder implements BuilderInterface {
...
+    const ctx = Context.create(settings);
+    const builderPtr = checkPtr(
+      getLib().c2pa_builder_from_context(ctx.ptr),
+      "Failed to create C2paBuilder",
     );
+    const stream = C2paStream.fromSource(asset);
+    try {
+      const newPtr = checkPtr(
+        getLib().c2pa_builder_with_archive(builderPtr, stream.ptr),
+        "Failed to load archive",
```
tmathern, line 253:
```suggestion
        "Failed to load Builder archive",
```

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -188,83 +280,102 @@ export class Builder implements BuilderInterface {
...
+    }
+  }
+
+  /**
+   * Async sign path used by signAsync()/signConfigAsync() — dispatches the
+   * native call via koffi's .async() (libuv threadpool) instead of calling
+   * it synchronously, matching the Neon binding's own async behavior for
+   * these methods.
+   */
+  private async _signWithAsync(
```
tmathern, line 334: Very similar to _signWith. Maybe there is a refactoring opportunity for the checks/initializations here at least?

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -188,83 +280,102 @@ export class Builder implements BuilderInterface {
...
+    input: SourceAsset,
+    output: DestinationAsset,
+  ): Promise<Buffer> {
+    const src = C2paStream.fromSource(input);
+    const dest = C2paStream.forDestination(output);
+    try {
+      const nativeSigner = signer.nativeSigner();
+      const manifestBytesOut = [null];
+      const size = checkIntAsync(
+        await callAsync<number | bigint>(
```
tmathern, line 345: Why not call a promisified version of `_signWith` at this point? Does koffi provide additional things that make it a better way to make this async? Or safer?

packages/c2pa-node/js-src/IdentityAssertion.ts
```diff
@@ -11,112 +11,230 @@
...
+  }
+
+  sigType(): string {
+    return this._sigType;
+  }
+
+  /**
+   * c2pa_identity_signer_create needs a real, already-built C2paSigner*
+   * for the identity side. A genuinely async JS callback can't be bridged
+   * into one — same root cause as CallbackSigner (see Signer.ts and
```
tmathern, line 121: Can this be worked around by having something sync in the C FFI? (May need a tokio runtime)?

packages/c2pa-node/js-src/IdentityAssertion.ts
```diff
@@ -11,112 +11,230 @@
...
-  ) {}
+  private _pendingAssertion?: IdentityAssertionBuilder;
+  private _combined?: NativeSigner;
 
-  getHandle(): NeonCallbackCredentialHolderHandle {
-    return this.callbackCredentialHolder;
+  private constructor(private _c2paSigner: HasNativeSigner) {}
+
+  /** `signer` is the opaque handle from a LocalSigner/CallbackSigner's
+   * getHandle() — matches this codebase's existing convention (see
```
tmathern, line 184: Maybe you could also create a dedicated type like `NativeHandle`, that wraps all those... ?

packages/c2pa-node/js-src/Signer.ts
```diff
@@ -11,104 +11,191 @@
...
+/** Internal contract Builder uses to reach the underlying native signer,
+ * regardless of which public signer class this is. */
+export interface HasNativeSigner {
+  nativeSigner(): NativeSigner;
+  /** See CallbackSigner.nativeSigner() — the real reason a sync callback
+   * bridge failed, when the native error alone wouldn't explain it. */
+  lastSyncError?(): Error | undefined;
+}
+
+function certsToPem(certs: Array<Buffer> | undefined): string {
```
tmathern, line 33: Add docs here, why is this conversion needed?

packages/c2pa-node/js-src/Signer.ts
```diff
@@ -11,104 +11,191 @@
...
-    );
-    return new LocalSigner(signer);
+    return new LocalSigner(algorithm, [certificate], privateKey, tsaUrl);
   }
 
+  /** Sign raw bytes directly with Node's built-in crypto (no native call —
+   * this is a standalone utility surface, separate from the manifest-signing
+   * pipeline in Builder.sign(), which uses the native signer via
+   * nativeSigner() instead). */
   sign(data: Buffer): Buffer {
```
tmathern, line 60: What is this API used for then? We should try to keep the signing logic, any signing logic (except when callbacks are passed in) in c2pa-rs.

packages/c2pa-node/js-src/Trustmark.ts
```diff
@@ -11,36 +11,28 @@
...
 
-import { getNeonBinary } from "./binary.js";
-import type {
-  TrustmarkInterface,
-  TrustmarkConfig,
-  NeonTrustmarkHandle,
-} from "./types.d.ts";
+import type { TrustmarkInterface, TrustmarkConfig } from "./types.d.ts";
+
+function notImplemented(): Error {
```
tmathern, line 16: Maybe instead we need to introduce abstractions (classes/interfaces/way to inject objects) so this could be implementable without that? Or work with a js implementation from https://github.com/adobe/trustmark/tree/main/js ?

packages/c2pa-node/README.md
```diff
@@ -635,12 +646,16 @@ const builder = Builder.new(urlSettings);
...
 
-### Build and use custom binary
-**Build Rust binary:**
-```bash
-pnpm run build:rust
-```
-**Use custom Rust binary:**
+### Build and use a custom native library
+
+On this branch, `C2PA_LIBRARY_PATH` points at a prebuilt `libc2pa_c` shared
```
tmathern, line 651: How will this be packaged and distributed on npm? gyp build steps?

tmathern (commented):
General notes:
- Something needs to call dispose at a point in time, this should go in docs or...
- You may need to consider introducing a finalization registry here (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry), which may help with cleanup. The question is how to design automated clean up if any here. It depends how much of the memory handling logic you want to expose to end users (if they free or the SDK does, I think it's better to let callers control when the memory is freed when possible... but it is easy to forget to do too.).
- For the missing API: please at least try to define input/output. Some may not even need to be C FFI, as the C FFI is kept small on purpose.

ale-adobe:
Closed as per discussion in the proposal. https://wiki.corp.adobe.com/x/kwLV7

PR #155, opened 2026-07-16
feat: Update for c2pa version 0.89.3

cdmurph32:


packages/c2pa-wasm/Cargo.toml
```diff
@@ -28,6 +28,13 @@ serde-wasm-bindgen = "0.6.5"
 serde = "1.0.219"
 serde_bytes = "0.11.19"
 
+# Enable the JS-backed randomness backend for both getrandom majors pulled in
+# transitively (0.3 via c2pa, 0.4 via lopdf/rand). Paired with the
+# getrandom_backend="wasm_js" cfg in .cargo/config.toml.
```
ale-adobe: I don't see this in the `.cargo/config.toml`. Did we forget to commit a change there?

cdmurph32: Good catch. That comment is just wrong.

.changeset/breezy-meteors-sit.md
```diff
@@ -0,0 +1,8 @@
+---
+'@contentauth/c2pa-wasm': patch
+'@contentauth/c2pa-node': patch
+'@contentauth/c2pa-types': patch
+'@contentauth/c2pa-web': patch
```
ale-adobe, line 5: I think we've typically bumped these as minor when updating the upstream `c2pa-rs` dependency.

ale-adobe (approved):

.changeset/breezy-meteors-sit.md
cdmurph32: I always select patch when there is no change in the API.

PR #157, opened 2026-07-17
fix: Fix node binary upload

cdmurph32:


tmathern (approved):

scouten-adobe (approved):

.github/workflows/release.yml
```diff
@@ -66,6 +65,19 @@ jobs:
         run: |
           echo '${{ needs.release.outputs.publishedPackages }}'
 
+      - name: Resolve published c2pa-node version
```
tmathern, line 68: Would the other packages need a similar exercise?

cdmurph32: I don't think so. This is just for the binary upload.

PR #161, opened 2026-07-23
feat: Add retry logic when fetching trust anchors

ale-adobe:
This PR adds basic exponential backoff retry mechanism when fetching resources in `settings.ts`, which is used for retrieving trust resources. This is a localized retry solution for the short-term, which somewhat mirrors the logic that exists elsewhere in Node. We will revisit this later as part of the overall improvements to the JS SDKs to have a more shared retry mechanism for re-use across the SDKs.

### DIFF
--- a/.changeset/wet-coats-doubt.md
+++ b/.changeset/wet-coats-doubt.md
@@ -1,0 +1,5 @@
+---
+'@contentauth/c2pa-web': patch
+---
+
+Add retry with exponential backoff for fetching trust resources.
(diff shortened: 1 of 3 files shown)


tmathern (approved):
Follow-up to consider: 429 may send a retry-after header too. You may want to parse that and honor it (if reasonable) or fail if the retry-after is very, very far away.

tmathern:
(Would be nice to have a generic retry function too, in the follow-up refactor. It looks like it is already going that way ;) ).

ale-adobe:
Added a way to respect the `retry-after` header if provided, with 30 seconds as the cutoff as being too long.

PR #162, opened 2026-07-23
feat: Include new experimental builder reduction methods

cdmurph32:
Incorporates [feat(sdk): add Builder action/ingredient removal API](https://github.com/contentauth/c2pa-rs/pull/2281/changes#top)

### DIFF
--- a/packages/c2pa-node/js-src/types.d.ts
+++ b/packages/c2pa-node/js-src/types.d.ts
@@ -344,20 +344,22 @@ export interface BuilderInterface {
   addRedaction(uri: string, reason: C2paReason): void;
 
   /**
-   * Retains only the actions for which `keep` returns true. **Experimental.**
+   * **Experimental.**
+   * Retains only the actions for which `keep` returns true.
    *
    * The inception action (`c2pa.created`/`c2pa.opened`) is always kept regardless of `keep`,
    * and is moved to index 0 if needed, so the manifest stays valid per the C2PA spec. Sets
    * `allActionsIncluded = false` when anything is removed. This does not touch ingredients —
    * call {@link filterIngredients} (`filterIngredients(() => false)` to drop all orphans)
    * afterwards if you also want to drop ingredients now orphaned by the removed actions.
-   * @param keep A predicate; the action is retained when it returns true.
+   * @param keep The action is retained when the predicate returns true.
    */
   filterActions(keep: (action: Action) => boolean): void;
 
   /**
+   * **Experimental.**
    * Retains ingredients, then rewrites positional ingredient references so linked actions
-   * stay valid. **Experimental.**
+   * stay valid.
    *
    * An ingredient is kept if it is referenced by a current action, is a `parentOf` ingredient,
    * or `rescue` returns true for it. `rescue` therefore only ever rescues an otherwise-orphaned
@@ -369,7 +371,7 @@ export interface BuilderInterface {
    * `manifest_data` parsed into a {@link ManifestStore}, or `null` when it has no embedded
    * manifest. This lets the predicate make provenance-aware decisions (e.g. "this ingredient's
    * chain contains AI") without re-reading the whole builder first.
-   * @param rescue A predicate that can rescue an otherwise-orphaned ingredient by returning true.
+   * @param rescue Can rescue an otherwise-orphaned ingredient by returning true.
    */
   filterIngredients(
     rescue: (
--- a/packages/c2pa-node/js-src/types.d.ts
+++ b/packages/c2pa-node/js-src/types.d.ts
@@ -14,6 +14,7 @@
 // This file contains types not included in @contentauth/c2pa-types and not directly applicable to the neon generated code (index.node.d.ts)
 import { Buffer } from "buffer";
 import type {
+  Action,
   BuilderIntent,
   C2paReason,
   Ingredient,
@@ -21,7 +22,7 @@ import type {
   ManifestStore,
 } from "@contentauth/c2pa-types";
 
-export type { C2paReason, Ingredient } from "@contentauth/c2pa-types";
+export type { Action, C2paReason, Ingredient } from "@contentauth/c2pa-types";
 
 /**
  * Describes the digital signature algorithms allowed by the C2PA spec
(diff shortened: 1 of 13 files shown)


packages/c2pa-node/js-src/types.d.ts
```diff
@@ -342,6 +343,41 @@ export interface BuilderInterface {
...
+   * `rescue` receives the ingredient and its `provenance`: the ingredient's embedded
+   * `manifest_data` parsed into a {@link ManifestStore}, or `null` when it has no embedded
+   * manifest. This lets the predicate make provenance-aware decisions (e.g. "this ingredient's
+   * chain contains AI") without re-reading the whole builder first.
+   * @param rescue A predicate that can rescue an otherwise-orphaned ingredient by returning true.
+   */
+  filterIngredients(
+    rescue: (
+      ingredient: Ingredient,
+      provenance: ManifestStore | null,
```
ale-adobe, line 379: Can we name the parameter `manifestStore` instead, for fewer mental jumps?

packages/c2pa-web/src/lib/builder.ts
```diff
@@ -266,6 +309,30 @@ function createBuilder(
...
+    async filterIngredients(rescue: (ingredient: Ingredient) => boolean) {
+      const definition: ManifestDefinition = await tx.builder_getDefinition(id);
+      const ingredients: Ingredient[] = definition.ingredients ?? [];
+      const indices = ingredients.reduce<number[]>((rescued, ingredient, i) => {
+        if (rescue(ingredient)) {
+          rescued.push(i);
+        }
+        return rescued;
+      }, []);
+      await tx.builder_filterIngredientsAt(id, indices);
```
ale-adobe, line 350: Why does Web have to figure out the indices first in the JS code and then pass those into the RPC, whereas Node can just pass the predicate and have Rust do that logic?

packages/c2pa-node/src/neon_builder.rs
```diff
@@ -613,6 +614,111 @@ impl NeonBuilder {
...
+    ///
+    /// If `keep` throws or returns a non-boolean, the exception is surfaced to the caller. Because
+    /// the underlying filter mutates in place, the builder may be left partially filtered when the
+    /// predicate throws midway; callers should discard it on error.
+    pub fn filter_actions(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+        let rt = runtime();
+        let this = cx.this::<JsBox<Self>>()?;
+        let keep = cx.argument::<JsFunction>(0)?;
+        let undefined = cx.undefined();
+        let mut builder = rt.block_on(async { this.builder.lock().await });
```
tmathern, line 634: Is it ok to block_on here and hold a lock? Can't this deadlock? 

I'm thinking especially if you access the builder in the predicate, which may be tempting to do in the future, something like: `builder.filterActions(a => builder.getManifestDefinition().assertions.length > 0)`. 
(I think the two builders will contend for the lock).

We should at least note if there is a risk here.

packages/c2pa-node/src/neon_builder.rs
```diff
@@ -613,6 +614,111 @@ impl NeonBuilder {
...
+
+        // The c2pa filter closure is `FnMut(&Action) -> bool` with no error channel, so capture a
+        // JS exception out-of-band and re-raise it after filtering. Once a throw is pending we
+        // stop invoking JS entirely (calling more JS APIs with an exception pending is unsound).
+        let mut pending: Option<neon::result::Throw> = None;
+        let filter_result = builder
+            .filter_actions(|action| {
+                if pending.is_some() {
```
tmathern, line 645: Once a throw is pending we stop invoking JS entirely: but then this stops checking here too. This might be fine if we consider that any error should stop processing, but it could also lead to side-effects since we wouldn't continue checking things (especially in the ingredients counterpart, for the "rescues").

packages/c2pa-wasm/src/wasm_builder.rs
```diff
@@ -162,6 +162,55 @@ impl WasmBuilder {
         Ok(())
     }
 
+    /// Retains only the actions at the given 0-based indices into the actions currently returned
+    /// by [`Self::get_definition`]'s `c2pa.actions` assertion.
+    ///
+    /// The inception action (`c2pa.created`/`c2pa.opened`) is always kept regardless of
+    /// `indices`, and is moved to index 0 if needed, so the manifest stays valid per the C2PA
+    /// spec. Sets `allActionsIncluded = false` when anything is removed. This does not touch
+    /// ingredients — call [`Self::filter_ingredients_at`] (with an empty list to drop all
```
tmathern: I would put this sentence last of the doc so it is easy to see, it implies a call order

packages/c2pa-node/js-src/types.d.ts
```diff
@@ -342,6 +343,41 @@ export interface BuilderInterface {
    */
   addRedaction(uri: string, reason: C2paReason): void;
 
+  /**
+   * Retains only the actions for which `keep` returns true. **Experimental.**
```
tmathern:
```suggestion
   * **Experimental.**
   Retains only the actions for which `keep` returns true. 
```

packages/c2pa-node/js-src/types.d.ts
```diff
@@ -342,6 +343,41 @@ export interface BuilderInterface {
...
+   * `allActionsIncluded = false` when anything is removed. This does not touch ingredients —
+   * call {@link filterIngredients} (`filterIngredients(() => false)` to drop all orphans)
+   * afterwards if you also want to drop ingredients now orphaned by the removed actions.
+   * @param keep A predicate; the action is retained when it returns true.
+   */
+  filterActions(keep: (action: Action) => boolean): void;
+
+  /**
+   * Retains ingredients, then rewrites positional ingredient references so linked actions
+   * stay valid. **Experimental.**
```
tmathern: I would put Experimental first in the comment

packages/c2pa-node/js-src/types.d.ts
tmathern: `provenance` is of type `ManifestStore`. Is that not good enough?
Because `provenance` is a better word here than leaking type names, in my opinion.

packages/c2pa-wasm/src/wasm_builder.rs
```diff
@@ -162,6 +162,55 @@ impl WasmBuilder {
...
+    /// builder lives in a worker and JS callbacks can't be invoked synchronously across that
+    /// boundary.
+    #[wasm_bindgen(js_name = filterActionsAt)]
+    pub fn filter_actions_at(&mut self, indices: Vec<u32>) -> Result<(), JsString> {
+        let indices: std::collections::HashSet<u32> = indices.into_iter().collect();
+        let mut i: u32 = 0;
+        self.builder
+            .filter_actions(|_action| {
+                let keep = indices.contains(&i);
+                i += 1;
```
tmathern, line 188: In a far-fetched theory, this could overflow. But then, this means we would have already gone through that many action. Hard to reach, but could happen.

packages/c2pa-wasm/src/wasm_builder.rs
```diff
@@ -162,6 +162,55 @@ impl WasmBuilder {
...
+    ///
+    /// See [`Self::filter_actions_at`] for why this takes indices rather than a predicate.
+    #[wasm_bindgen(js_name = filterIngredientsAt)]
+    pub fn filter_ingredients_at(&mut self, indices: Vec<u32>) -> Result<(), JsString> {
+        let indices: std::collections::HashSet<u32> = indices.into_iter().collect();
+        let mut i: u32 = 0;
+        self.builder
+            .filter_ingredients(|_ingredient| {
+                let rescue = indices.contains(&i);
+                i += 1;
```
tmathern, line 210: Could in theory also overflow (but I think that given we create a Reader per ingredient, there would be other issues before if this managed to get at such a high value).

packages/c2pa-web/src/lib/builder.ts
```diff
@@ -104,6 +105,34 @@ export interface Builder {
    */
   addRedaction: (uri: string, reason: C2paReason) => Promise<void>;
 
+  /**
+   * Retains only the actions for which `keep` returns true. **Experimental.**
```
tmathern: Experimental on top of comment?

packages/c2pa-web/src/lib/builder.ts
```diff
@@ -104,6 +105,34 @@ export interface Builder {
...
+   *
+   * The inception action (`c2pa.created`/`c2pa.opened`) is always kept regardless of `keep`,
+   * and is moved to index 0 if needed, so the manifest stays valid per the C2PA spec. Sets
+   * `allActionsIncluded = false` when anything is removed. This does not touch ingredients —
+   * call {@link Builder.filterIngredients} (`filterIngredients(() => false)` to drop all
+   * orphans) afterwards if you also want to drop ingredients now orphaned by the removed
+   * actions.
+   *
+   * @param keep A predicate; the action is retained when it returns true.
```
tmathern:
```suggestion
   * @param keep The action is retained when the predicate returns true.
```
(So it doesn't read like "keep a predicate etc...").

packages/c2pa-node/src/neon_builder.rs
tmathern: (That risk I think is smaller in the ingredients counterpart, I would be less tempted to do such things there).

tmathern (approved):

packages/c2pa-node/js-src/types.d.ts
ale-adobe: I agree that `provenance` captures the concept better, but I don't think we're really leaking type names in this case -- we're already importing the `ManifestStore` type throughout both SDKs. 🤷

packages/c2pa-web/src/lib/builder.ts
cdmurph32: Because filter_actions is sync. If we made it async and matched the node workflow exactly, there would be a round-trip on every ingredient.

packages/c2pa-node/src/neon_builder.rs
cdmurph32: Yes. I'll add a comment.

PR #168, opened 2026-07-31
refactor: Initial Settings, retry, and fetch refactor

ale-adobe:
This initial refactor is centered around the introduction of a new package: `c2pa-utilities`, and unifies `Settings`, retry, and fetch logic shared between `c2pa-web` and `c2pa-node`.

Across the board, no new functionality was added. Types or functions maintain their current behavior, but are widened in some cases, due to the unification of shapes across Web and Node, in order to maintain support for both. This is not necessarily the final shape; this just helps put everything in one place so that it will be easier to iterate on moving forward.

### New additions
The new `c2pa-utilities` package contains runtime-agnostic logic that `c2pa-web` and `c2pa-node` previously implemented independently (and had drifted on):
- `Settings`/`TrustSettings`/`CawgTrustSettings`/`VerifySettings`/`BuilderSettings`: now one shared shape, mirroring structs in `c2pa-rs`
- Functions for handling settings resolution and serialization: `resolveSettings`, `createTrustSettings`/`createCawgTrustSettings`/`createVerifySettings`, `mergeSettings`, `settingsToJson`, `snakeCaseify`, `loadSettingsFromUrl`
- Functions for trust-anchor fetching: `resolveTrustSettings`

Additionally, this package houses generic HTTP retry mechanics that was also previously implemented by both packages. The new `fetchWithRetry.ts` module implements fetch-with-retry with exponential backoff and handles `Retry-After` headers along with configurable options, which will allow the Web and Node packages to inject their own runtime-specific behaviors.

### Removals
- `c2pa-web`
  - `lib/settings.ts` removed entirely; re-exports `c2pa-utilities` directly from `index.ts` for client convenience
- `c2pa-node`
  - `Settings.ts` reduced to just `loadSettingsFromFile` (unique to Node thanks to its file system access), plus a re-export of everything else via its respective `index.ts` for client convenience.
  - `TrustConfig`/`VerifyConfig`/`SettingsContext` kept as `@deprecated` type aliases for backward compatibility, but otherwise are just re-exposing the unified types now defined in `c2pa-utilities`

### Notable behavior changes / new capabilities
- `mergeSettings` now deep-merges nested fields (e.g. `builder.thumbnail`) instead of overwriting whole sections
- `c2pa-node` gains support for nested arrays for trust-anchor fields (previously only a web feature) and a new opt-in `resolveTrustAnchors()` for validated fetching
- Fixed a bug in `snakeCaseify`: array values were being flattened into `{"0": ..., "1": ...}` objects instead of preserved as arrays

### Testing
- Added new test cases for parity and edge cases, including network errors, `Retry-After` variants, non-`Error` rejections, size-cap enforcement, PEM validation across `settings.spec.ts` and the new `fetchWithRetry.spec.ts`

### Follow-up work
Several smaller PRs will be coming up containing their own bite-sized standalone changes, targeting this one as the base branch:
- Unification of format allowlists (configurable)
- Unification of size guards and limitations when fetching (configurable)
- Unification of retry mechanisms (configurable)
- Unification of `SigningAlg` and other shared enums


packages/c2pa-node/js-src/index.ts
```diff
@@ -23,3 +23,4 @@ export {
 export { Trustmark } from "./Trustmark.js";
 export { isActionsAssertion } from "./assertions.js";
 export * from "./Settings.js";
+export * from '@contentauth/c2pa-utilities';
```
ale-adobe, line 26: Re-exporting from utilities for client convenience.

packages/c2pa-node/js-src/Settings.spec.ts
```diff
@@ -4,230 +4,13 @@
...
 import * as fs from "fs-extra";
 import * as path from "path";
 import * as os from "os";
-import { vi } from "vitest";
 
-// Mock node-fetch
-vi.mock("node-fetch", () => ({
-  default: vi.fn(),
-}));
+import { loadSettingsFromFile } from "./Settings.js";
```
ale-adobe, line 11: Tests are consolidated and moved into the `settings.spec.ts` in `c2pa-utilities`. Only the one Node-specific function is tested here.

packages/c2pa-node/js-src/types.d.ts
```diff
@@ -195,7 +195,9 @@ export interface HashedUri {
 export type C2paSettings = string | object;
 
 export interface BuilderInterface {
-  /** An intent lets the API know what kind of manifest to create.
+  /**
```
ale-adobe, line 198: Small formatting adjustment.

packages/c2pa-utilities/src/fetchWithRetry.ts
```diff
@@ -0,0 +1,164 @@
...
+ * Options to configure {@link fetchWithRetry}. Retry count and backoff timing are fixed
+ * internally for now — only the response-size cap is currently overridable.
+ */
+export interface FetchWithRetryOptions {
+  /**
+   * Maximum allowed response size, in bytes.
+   * Defaults to {@link DEFAULT_MAX_RESPONSE_BYTES} when omitted.
+   */
+  maxResponseBytes?: number;
+}
```
ale-adobe: This will be expanded on in follow-up PRs to allow for greater configuration by the Web and Node SDKs independently.

packages/c2pa-web/src/lib/c2pa.ts
```diff
@@ -51,6 +51,19 @@ export interface C2paSdk {
   dispose: () => void;
 }
 
+/**
```
ale-adobe, line 54: This comment was leftover in `c2pa-web/src/index.ts`. Not sure when it got lost, but it's meant to be attached to this function, so moving it over here now.

packages/c2pa-web/src/common.ts
```diff
@@ -24,13 +24,5 @@ export {
   READER_SUPPORTED_FORMATS
 } from './lib/supportedFormats.js';
 
-export type {
```
ale-adobe, line 27: These are now re-exported from `index.ts`.

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,436 @@
+/**
```
ale-adobe, line 1: This file unifies Settings-related logic from both Web and Node SDKs, in some cases widening types and function signatures to support both, but without changing behavior. It also re-organizes code into more logical chunks.

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,436 @@
...
+  /**
+   * Settings for automatic thumbnail generation.
+   */
+  thumbnail?: {
+    /**
+     * Whether to automatically generate a thumbnail for the asset being built, if possible.
+     */
+    enabled?: boolean;
+  };
+}
```
ale-adobe, line 147: This was union'd between the two packages. One had `generateC2paArchive` as an optional setting, and the other had `thumbnail` as an optional setting. They now both exist here in the new `BuilderSettings` as optional.

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,436 @@
...
+ * @param baseSettings Settings established at SDK initialization time.
+ * @param overrideSettings Optional override settings. Keys present in overrideSettings win over keys in baseSettings.
+ * @param options Optional per-call overrides, such as the trust-anchor response size cap.
+ * @returns A JSON-serialized string containing all resolved settings values, ready to be consumed by c2pa-rs.
+ * Returns undefined when neither argument is provided.
+ */
+export async function resolveSettings(
+  baseSettings: Settings | undefined,
+  overrideSettings: Settings | undefined,
+  options?: FetchWithRetryOptions
```
ale-adobe, line 176: These options could be generalized going forward, rather than being specific to `FetchWithRetryOptions`. However, at the moment, the only options that are passed in are specific to fetch-with-retry configuration, so I'm opting to name it as such for now.

packages/c2pa-utilities/src/fetchWithRetry.ts
```diff
@@ -0,0 +1,164 @@
+/**
```
ale-adobe, line 1: Considering naming this file just `fetch.ts`.

packages/c2pa-node/js-src/types.d.ts
```diff
@@ -478,62 +480,26 @@ export interface TrustmarkConfig {
...
-  /** User-provided trust anchors (PEM format or base64-encoded certificate hashes) */
-  userAnchors?: string;
-  /** Trust anchors for validation (PEM format or base64-encoded certificate hashes) */
-  trustAnchors?: string;
-  /** Trust configuration file path */
-  trustConfig?: string;
-  /** Allowed list of certificates (PEM format or base64-encoded certificate hashes) */
-  allowedList?: string;
-}
+export type TrustConfig = _CawgTrustSettings;
```
tmathern: I don't think that deprecation note is correct. Trust was not only for CAWG, it is the general trust (as in: is the signing manifest trusted?).
(Except if it means "usual" trust + CAWG trust are merged together in one type, but that could use a clearer deprecation note too).

packages/c2pa-utilities/src/fetchWithRetry.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+          new HttpResponse(null, { status: 429, statusText: 'Too Many Requests' })
+      )
+    );
+
+    await expect(fetchWithRetry('http://always429')).rejects.toThrow(
+      'Failed to fetch http://always429: 429'
+    );
+  });
+
+  test('respects a reasonable numeric Retry-After header and retries', async () => {
```
tmathern: Instead of `reasonable`, test it does not exceed a maximum and state the max or the constant in which the max value is defined.

packages/c2pa-utilities/src/fetchWithRetry.ts
```diff
@@ -0,0 +1,164 @@
...
+        continue;
+      }
+      const message = e instanceof Error ? e.message : String(e);
+      throw new Error(`Network error fetching ${url}: ${message}`, {
+        cause: e
+      });
+    }
+
+    if (!res.ok) {
+      const retryable = res.status === 429 || res.status >= 500;
```
tmathern: In follow-ups, consider making also the retryable codes configurable. 
Some server codes in some setups might be considered non retryable (e.g. 503, sometimes).

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,414 @@
...
+ *
+ * NOTICE: Adobe permits you to use, modify, and distribute this file in
+ * accordance with the terms of the Adobe license agreement accompanying
+ * it.
+ */
+
+import { merge } from 'ts-deepmerge';
+import { fetchWithRetry, type FetchWithRetryOptions } from './fetchWithRetry.js';
+
+// =============================================================================
```
tmathern: Let's keep the separator length to a smaller length, shall we? This breaks side-by-side UI review in Github on laptop, and makes code difficult to read on smaller laptop screens too.

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,414 @@
...
+// Types
+// =============================================================================
+
+/**
+ * Settings configuration for C2PA operations.
+ *
+ * Encapsulates settings and configuration options for Reader and Builder operations.
+ * It provides a flexible way to configure SDK behavior including the verification configuration,
```
tmathern: You can drop or summarize that line, it basically says agains it's settings.

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,414 @@
...
+ * trust configuration, and builder options.
+ *
+ * This shape mirrors `c2pa-rs`'s own `Settings` struct.
```
tmathern: That line can be removed too

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,414 @@
...
+}
+
+export interface BuilderSettings {
+  /**
+   * Whether to generate a C2PA archive (instead of zip) when writing the manifest builder.
+   *
+   * Note: `c2pa-rs` is deprecating the zip archive path — this setting is expected to be
+   * removed in a future release and should always be left `true`.
+   */
+  generateC2paArchive?: boolean;
```
tmathern: Note that that settings is obsolete. It is "true" by default now, and if you set it to false, the SDK will tell you the feature is not implemented since ZIP support was dropped. 
Remove it, or set it to true (not updatable) to avoid confusion.

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,414 @@
...
+  };
+}
+
+// =============================================================================
+// Defaults
+// =============================================================================
+
+const DEFAULT_SETTINGS: Settings = {
+  builder: {
+    generateC2paArchive: true
```
tmathern, line 155: True by default in code SDK already now.

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,414 @@
...
+  }
+};
+
+// =============================================================================
+// Top-level pipeline
+// =============================================================================
+
+/**
+ * Resolves settings by merging override settings on top of base settings, resolving any embedded
+ * trust list URLs on top of those, and then finally serializing the result for consumption by c2pa-rs.
```
tmathern:
```suggestion
 * trust list URLs on top of those, and then finally serializing the result for consumption by the core native SDK
```

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,414 @@
...
+  | undefined
+  | SettingsObjectType
+  | SettingsValue[];
+type SettingsObjectType = {
+  [k: string]: SettingsValue;
+};
+
+/**
+ * Recursively converts an object's camelCase keys to snake_case, matching the format
+ * `c2pa-rs` expects for settings JSON/TOML. Arrays are preserved as arrays — only their
```
tmathern: We should not support TOML here. The core SDK itself is trying to refocus on using JSON too.

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,414 @@
...
+ * object elements (if any) get their keys snake-cased — since `typeof [] === 'object'` in
+ * JS and a naive `Object.entries`-based recursion would otherwise flatten an array into a
+ * `{"0": ..., "1": ...}` object instead of a JSON array.
+ */
+export function snakeCaseify(object: SettingsObjectType): SettingsObjectType {
```
tmathern: That's a lot of boilerplate, which is potentially generic and reusable too. Move it into another file.

tmathern (changes requested):
- Please clarify the deprecation note in the types, it is confusing.
- Also review comments in settings.

packages/c2pa-utilities/src/fetchWithRetry.ts
ale-adobe: Yes, I have a PR in the works for this already.

packages/c2pa-utilities/src/settings.ts
ale-adobe: Good to know, will update.

packages/c2pa-utilities/src/settings.ts
ale-adobe: Great!

packages/c2pa-utilities/src/settings.ts
ale-adobe: Good callout, will move out to further focus this file.

packages/c2pa-node/js-src/types.d.ts
ale-adobe: Yes, it is because they are merged together into one type, and `CawgTrustSettings` (inherited from `c2pa-web`) encapsulates all of them. So the `CawgTrustSettings` now in the shared `c2pa-utilities` is actually what `c2pa-node` used to refer to as its `TrustConfig`. I'll update the note.

cdmurph32 (changes requested):
My only concern here is that this doesn't break anything. Could you please add tests in the c2pa-js and c2pa-node repositories to ensure the settings are applied?

packages/c2pa-node/js-src/types.d.ts
tmathern: Why not just keep it named trust? Cawg is a set of trust. Because as a caller, it would not occur to me that this is also for something else than Cawg trust. Cawg is an "additional" type of trust here.

ale-adobe: I didn't want to change the existing shape as used by `c2pa-web` currently: https://github.com/contentauth/c2pa-js/blob/main/packages/c2pa-web/src/lib/settings.ts#L88

That said, I'm not opposed to just collapsing it all into a single `TrustSettings` interface that contains all 5 optional fields. I believe it all goes down to the native library the same way anyways.

tmathern: Let's keep it generic in naming if possible

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,358 @@
...
+   * Possible values are: the text content of a end-entity cert file, a URL to fetch a end-entity cert file from, or an array of URLs that will be fetched and concatenated.
+   */
+  allowedList?: string | string[];
+}
+
+export interface CawgTrustSettings extends TrustSettings {
+  /**
+   * Enable CAWG trust validation. The default value is "true."
+   */
+  verifyTrustList?: boolean;
```
tmathern, line 92: We don't want it off usually, but shouldn't this specific setting be in the base class too?

.changeset/fresh-kiwis-doubt.md
```diff
@@ -0,0 +1,7 @@
+---
+'@contentauth/c2pa-utilities': minor
```
tmathern, line 2: They all should be minor, I suspect changes will propagate and be potentially visible by users too.

packages/c2pa-utilities/src/fetchWithRetry.ts
```diff
@@ -0,0 +1,242 @@
...
+  options?: FetchWithRetryOptions
+): Promise<string> {
+  const maxResponseBytes = options?.maxResponseBytes ?? DEFAULT_MAX_RESPONSE_BYTES;
+
+  const res = await fetchWithRetryRaw(url, undefined, options);
+  const text = await res.text();
+
+  if (text.length > maxResponseBytes) {
+    throw new Error(
+      `Response from ${url} is too large. Max size is ${maxResponseBytes} bytes.`
```
tmathern, line 265: It is already in buffer though, since `res.text();` has been called. You can try to rely somewhat on the content-length honor system isntead.

packages/c2pa-utilities/src/settings.ts
```diff
@@ -0,0 +1,358 @@
...
+// =================================
+
+/**
+ * Load settings from a URL.
+ *
+ * @param url The URL to fetch the settings from
+ * @returns Settings as a string
+ */
+export async function loadSettingsFromUrl(url: string): Promise<string> {
+  const res = await fetch(url);
```
tmathern: No retry here?

packages/c2pa-utilities/src/fetchWithRetry.ts
```diff
@@ -0,0 +1,242 @@
...
+ * maximum response size. A thin convenience wrapper around {@link fetchWithRetryRaw} for the
+ * common GET-and-read-as-text case; use {@link fetchWithRetryRaw} directly for other methods,
+ * request bodies, or response handling.
+ *
+ * @param url The URL to fetch.
+ * @param options Options for configuring the fetch.
+ * @returns The response body.
+ */
+export async function fetchWithRetry(
+  url: string,
```
tmathern, line 247: Since url gets in as string, maybe check also it is a valid URL beforehand?

tmathern (commented):
(Probably also update the PR description, I got no search hit on `resolveTrustAnchors` in this changeset, other mentioned functions are here).

.changeset/fresh-kiwis-doubt.md
ale-adobe: When this lands, the bot should end up bumping them all as minor because `c2pa-node` and `c2pa-web` depend on `c2pa-utilities`, and since `c2pa-utilities` is a minor change, it would propagate to those as well. I'll update it here to be more explicit about it though.

tmathern (approved):

packages/c2pa-node/js-src/types.d.ts
ale-adobe: Collapsed into just `TrustSettings`.

packages/c2pa-utilities/src/settings.ts
ale-adobe: Yes, collapsed and combined into one class.

packages/c2pa-utilities/src/settings.ts
ale-adobe: Added retry here, using `fetchWithRetryRaw`.

packages/c2pa-utilities/src/fetchWithRetry.ts
ale-adobe: Checking it is a valid URL by constructing a new `URL()` now.

packages/c2pa-utilities/src/fetchWithRetry.ts
ale-adobe: Added a check for content-length, but still using `res.text()` as a fallback since that header is optional.

cdmurph32 (approved):
approved

PR #169, opened 2026-08-03
refactor: shared asset size validation

ale-adobe:
This PR pulls out asset size validation from `c2pa-web` into a shared utility library in `c2pa-utilities`, and then adds that validation check to `c2pa-node`'s `Reader` factories.

`c2pa-web` maintains its existing 1 GB size limit, while `c2pa-node` gains a new 10 GB size limit. These are hard-coded at the moment in the source code and cannot be configured by clients. In the future, I intend to make these configurable once we have a more formalized concept of a `Context` and we can better support global settings per SDK instance in a way that is easily shared between both packages.


packages/c2pa-web/src/lib/reader.spec.ts
```diff
@@ -58,6 +60,29 @@ describe('reader', () => {
         expect(reader).toBeNull();
       });
 
+      test('should throw UnsupportedFormatError for an unsupported format', async ({
```
ale-adobe, line 63: Adding some missing unit test coverage for error handling.

packages/c2pa-node/js-src/assetSize.ts
```diff
@@ -0,0 +1,35 @@
...
+// implied. See the LICENSE-MIT and LICENSE-APACHE files for the
+// specific language governing permissions and limitations under
+// each license.
+
+import * as fs from "fs-extra";
+import { validateAssetSize } from "@contentauth/c2pa-utilities";
+import type { SourceAsset } from "./types.d.ts";
+
+// c2pa-node runs server-side, so it can reasonably support larger assets than
+// c2pa-web, which runs in a browser tab with real memory and UX constraints.
```
tmathern: A server also has real memory :)

I would rather emphasize the limitations, that some browsers may be more resource constrained, to justify this.

packages/c2pa-utilities/src/assetSize.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * This file implements the shared Reader asset size check.
+ * 
+ * The size threshold is always supplied by the caller, since each platform
+ * (`c2pa-web` or `c2pa-node`) decides its own limit explicitly rather than
+ * inheriting a hardcoded default from here.
+ */
+
+export class AssetTooLargeError extends Error {
+  constructor(sizeInBytes: number, maxSizeInBytes: number) {
```
tmathern, line 43: What is the defined behavior if size is 0? Asking because  in some cases it means "no limit", so at least the comment needs to explain it. Sometimes it means "reset to default". Let's pick a behavior and make it clear here.

packages/c2pa-node/js-src/assetSize.ts
```diff
@@ -0,0 +1,35 @@
...
+export const MAX_SIZE_IN_BYTES = 10 * 10 ** 9; // 10 GB
+
+/**
+ * Validates a {@link SourceAsset}'s size before it's handed to the native reader.
+ */
+export async function validateSourceAssetSize(
+  asset: SourceAsset,
+): Promise<void> {
+  if ("buffer" in asset) {
+    validateAssetSize(asset.buffer.length, MAX_SIZE_IN_BYTES);
```
tmathern, line 37: If you can check buffer length here, it means it is allocated. Might need a note in the docs.

tmathern (commented):
- If fragmented media are supported, make sure also all fragments are checked as they go in (not just the first one).
- This is a minor change (new error code exported in the API), not a patch.

packages/c2pa-node/js-src/assetSize.ts
ale-adobe: Updated in next commit.

packages/c2pa-utilities/src/assetSize.ts
ale-adobe: Made the logic here more robust overall. Anything 0 or non-sensical is now converted to a default of 1 GB instead.

packages/c2pa-node/js-src/assetSize.ts
ale-adobe: Added a note, encouraging consumers to use `FileAsset` if concerned about memory spikes.

packages/c2pa-web/src/lib/reader.ts
```diff
@@ -164,9 +166,8 @@ export function createReaderFactory(worker: WorkerManager, settings?: Settings):
         throw new UnsupportedFormatError(format);
       }
 
-      if (init.size > MAX_SIZE_IN_BYTES) {
-        throw new AssetTooLargeError(init.size);
-      }
+      validateAssetSize(init.size, MAX_SIZE_IN_BYTES);
+      validateAssetSize(fragment.size, MAX_SIZE_IN_BYTES);
```
ale-adobe, line 170: We were never checking the fragment size before, but we are now!

packages/c2pa-node/js-src/assetSize.ts
```diff
@@ -0,0 +1,40 @@
...
+// specific language governing permissions and limitations under
+// each license.
+
+import * as fs from "fs-extra";
+import { validateAssetSize } from "@contentauth/c2pa-utilities";
+import type { SourceAsset } from "./types.d.ts";
+
+// c2pa-node runs server-side, so it can reasonably support larger assets than c2pa-web,
+// and may be less resource-constrained in comparison.
+export const MAX_SIZE_IN_BYTES = 10 * 10 ** 9; // 10 GB
```
tmathern, line 20: Is that export properly propagated to this package's index.ts?
Asking because there is this one too: https://github.com/contentauth/c2pa-js/blob/0431e1b7122bf0a96179a0dda0e7693cfe158aab/packages/c2pa-node/js-src/index.ts#L26

packages/c2pa-node/js-src/Reader.ts
```diff
@@ -43,6 +44,7 @@ export class Reader implements ReaderInterface {
   }
 
   static async fromAsset(asset: SourceAsset, settings?: C2paSettings): Promise<Reader | null> {
+    await validateSourceAssetSize(asset);
```
tmathern, line 47: Note this changes the potential reported errors on wrong or missing paths too or wrong access rights (due to the fs.stat).

packages/c2pa-node/js-src/Reader.ts
```diff
@@ -54,6 +56,7 @@ export class Reader implements ReaderInterface {
     asset: SourceAsset,
     settings?: C2paSettings,
   ): Promise<Reader> {
+    await validateSourceAssetSize(asset);
```
tmathern, line 59: Note this changes the potential reported errors on wrong or missing paths too or wrong access rights (due to the fs.stat).

packages/c2pa-web/README.md
```diff
@@ -81,6 +81,8 @@ console.log(manifestStore);
 await reader.free();
 ```
 
+`Reader.fromBlob` and `Reader.fromBlobFragment` reject assets larger than 1 GB, throwing `AssetTooLargeError`.
```
tmathern: You mean per fragment? Because they still validate separately in the code, not together (their sizes aren't summed):
```
validateAssetSize(init.size, MAX_SIZE_IN_BYTES);
validateAssetSize(fragment.size, MAX_SIZE_IN_BYTES);
```

packages/c2pa-node/js-src/assetSize.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Validates a {@link SourceAsset}'s size before it's handed to the native reader.
+ *
+ * For a `FileAsset`, this rejects an oversized file before it's read into memory.
+ * For a `SourceBufferAsset`, the buffer is already fully allocated by the time this runs,
+ * so the memory cost isn't prevented, only reader construction is. Prefer `path`
+ * for large or untrusted assets.
```
tmathern, line 28: Note also that this won't prevent a file to potentially grow above the limit after we checked, since we don't repeatedly check (it's fine, just a known limitation, can't continuously keep rechecking it!).

tmathern (approved):

packages/c2pa-web/README.md
ale-adobe: Yes; missed that in the docs. Will update.

packages/c2pa-node/js-src/assetSize.ts
ale-adobe: Yes, true. But I think this is good enough for now.

packages/c2pa-node/js-src/Reader.ts
ale-adobe: Updated docs to indicate that this is a possibility, without trying to enumerate them all.

packages/c2pa-node/js-src/Reader.ts
ale-adobe: Updated docs to indicate that this is a possibility, without trying to enumerate them all.

packages/c2pa-node/js-src/assetSize.ts
ale-adobe: No, and it's not intended to be. The default is used internally for now, but this may change later when we make this more configurable by clients of the SDK. At the moment, this is only being exported for testing purposes, but we can't differentiate between the two in JS.

cdmurph32 (approved):
LGTM

PR #177, opened 2026-08-05
feat: updateActions

cdmurph32:
Allow for modifying c2pa Actions in place.

### DIFF
--- a/packages/c2pa-node/js-src/Builder.spec.ts
+++ b/packages/c2pa-node/js-src/Builder.spec.ts
@@ -1218,7 +1218,7 @@ describe("Builder", () => {
           action.action === "c2pa.created"
             ? {
                 ...action,
-                parameters: { ...action.parameters, genAiId: "fixed-id" },
+                parameters: { ...action.parameters, patchMe: "patch-test" },
               }
             : action,
         ),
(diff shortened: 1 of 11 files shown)


packages/c2pa-node/js-src/Builder.spec.ts
```diff
@@ -1209,5 +1209,100 @@ describe("Builder", () => {
...
+
+    it("updateActions patches a parameter on an existing action", () => {
+      const builder = Builder.withJson(actionsManifest());
+
+      builder.updateActions((actions) =>
+        actions.map((action) =>
+          action.action === "c2pa.created"
+            ? {
+                ...action,
+                parameters: { ...action.parameters, genAiId: "fixed-id" },
```
tmathern: (We should pick a generic field name to patch, like `patch-me` or `patch-test` (with the proper casing)).

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -814,6 +814,164 @@ describe('builder', () => {
...
+            }
+          ]
+        } as unknown as ManifestDefinition);
+
+        await builder.updateActions((actions) =>
+          actions.map((action) =>
+            action.action === 'c2pa.created'
+              ? {
+                  ...action,
+                  parameters: { ...action.parameters, genAiId: 'fixed-id' }
```
tmathern: (Suggestion to pick a generic name for the second field here too)

packages/c2pa-node/src/neon_builder.rs
```diff
@@ -676,6 +676,56 @@ impl NeonBuilder {
...
+
+        builder
+            .definition
+            .assertions
+            .retain(|a| !a.label.starts_with(Actions::LABEL));
+
+        let mut updated = existing.unwrap_or_else(Actions::new);
+        updated.actions = updated_actions;
+        builder
+            .add_assertion(Actions::LABEL_VERSIONED, &updated)
```
tmathern: In c2pa-rs, this happens on add_assertion:

https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/builder.rs#L870-L874

`created` resets to false, `kind` to None.

Is this fine here? Maybe it needs to be moved somehow instead of deatched-reattached? (Is there a public type we could misuse?).

tmathern: (Especially, since in filter_actions, in the Builder: https://github.com/contentauth/c2pa-rs/blob/c2pa-v0.90.5/sdk/src/builder.rs#L1043-L1044).

packages/c2pa-node/src/neon_builder.rs
```diff
@@ -676,6 +676,56 @@ impl NeonBuilder {
...
+        let result = transform.call(&mut cx, undefined, [js_actions])?;
+        let updated_actions: Vec<Action> = neon_serde4::from_value(&mut cx, result)
+            .or_else(|err| cx.throw_error(err.to_string()))?;
+
+        builder
+            .definition
+            .assertions
+            .retain(|a| !a.label.starts_with(Actions::LABEL));
+
+        let mut updated = existing.unwrap_or_else(Actions::new);
```
tmathern: The updated actions are reattached as LABEL_VERSIONED.

If the actions where v1 LABEL, can we just migrate them like that? Reattach v1 actions to v2? E.g. this might change validation rules they are subject to (like presence of an opened or created action as first action in the actions array).

Asking because of the distinction between those tw: https://github.com/contentauth/c2pa-rs/blob/c2pa-v0.90.5/sdk/src/assertions/actions.rs#L912-L913

Also here, add action bothers about the original_label too:
https://github.com/contentauth/c2pa-rs/blob/c2pa-v0.90.5/sdk/src/builder.rs#L937-L953

packages/c2pa-web/src/lib/builder.ts
```diff
@@ -161,6 +161,16 @@ export interface Builder {
...
 
+  /**
+   * Replaces the actions in the `c2pa.actions`/`c2pa.actions.v2` assertion.
+   *
+   * `transform` receives the current actions (same enumeration as {@link Builder.filterActions})
+   * and returns the replacement list. Other properties are preserved as-is.
+   *
+   * @param transform Receives the current actions and returns the full replacement list.
+   */
+  updateActions: (transform: (actions: Action[]) => Action[]) => Promise<void>;
```
tmathern, line 181: That API also should have a warning about the transform: transform can do anything, it could "break" an actions array by, for instance, removing the inception action (or moving it to another position).

packages/c2pa-node/src/neon_builder.rs
```diff
@@ -676,6 +676,56 @@ impl NeonBuilder {
...
+        let undefined = cx.undefined();
+        let mut builder = rt.block_on(async { this.builder.lock().await });
+
+        let existing = builder
+            .definition
+            .assertions
+            .iter()
+            .find(|a| a.label.starts_with(Actions::LABEL))
+            .and_then(|a| serde_json::to_value(&a.data).ok())
+            .and_then(|v| serde_json::from_value::<Actions>(v).ok());
```
tmathern: The `ok()` here won't surface exceptions (and swallow it), this does not match the comment/intended behavior:
```
/// If `transform` throws or its return value doesn't deserialize into an action list, the
    /// exception is surfaced to the caller and the builder is left unchanged.
```
(Also when swallowing the errors, I am not sure this wouldn't continue with "wring" actions that are then wrongly reattached, if not empty).

tmathern (commented):
(I left comments on the Rust at node-side of things, but they likely apply to the Rust on the wasm side too since they are quite similar).

tmathern:
@cdmurph32 One thing I pondered: We could also keep the different kind of action assertions if there are multiple. That removes the troubles of collapsing into one, and possibly upgrading (so the manifest stays valid). What would you think of something like that: https://github.com/contentauth/c2pa-js/pull/179 ? (I started from this branch and built in review comments too).

Just an idea. PR can be closed without merging if going the wrong direction.

tmathern (approved):

PR #179, opened 2026-08-06
fix: Transform, but keep different actions assertions if multiple are present

tmathern:


cdmurph32 (approved):
LGTM. This is more correct than what I had.

PR #186, opened 2026-08-12
docs: Retitle READMEs to align with other repos, general copy edits

crandmck:
Although all the repo names will remain c2pa-*, I'm changing the README h1's to be CAI * library to help clarify that they are from CAI and *not* C2PA.

The general copy edit is to help align these docs with the style conventions used everywhere else.


ale-adobe (approved):
Thanks!

PR #187, opened 2026-08-12
chore: bundle ts-deepmerge for c2pa-web

ale-adobe:
Update `c2pa-web`'s vite config so that it bundles `ts-deepmerge` so that it can be properly consumed by other apps like our testbed. `c2pa-web` used to import `ts-deepmerge` directly, but that logic has since moved into `c2pa-utilities`. There is no longer any direct relationship between `c2pa-web` and `ts-deepmerge`, so it should be removed from the external list.


cdmurph32 (approved):
Approved. Thank you

PR #188, opened 2026-08-12
feat: Add isRetryableError and fetch options for FetchWithRetryOptions; do not retry on AbortError

ale-adobe:
Adds new options to `FetchWithRetryOptions`:
- `isRetryableError`: optional predicate accepting an error code and returning whether the code is retryable
- `fetch`: optional function matching the shape of `fetch` to allow overrides of fetch behavior, and allow injection of fetch functionality such as those provided through `wretch`

Also, do not allow retry on `AbortError`s.

### DIFF
--- a/packages/c2pa-utilities/src/fetchWithRetry.spec.ts
+++ b/packages/c2pa-utilities/src/fetchWithRetry.spec.ts
@@ -368,6 +368,33 @@ describe('fetchWithRetry', () => {
     const result = await fetchWithRetry('http://retryAfterOn503');
     expect(result).toBe('resolved after 503 retry-after');
   });
+
+  test('honors a custom isRetryableError, refusing to retry a network error retried by default', async () => {
+    const fetchMock = vi.fn().mockRejectedValue(new Error('boom'));
+    vi.stubGlobal('fetch', fetchMock);
+
+    await expect(
+      fetchWithRetry('http://customIsRetryableError', {
+        isRetryableError: () => false
+      })
+    ).rejects.toThrow('Network error fetching http://customIsRetryableError: boom');
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+  });
+
+  test('never retries an AbortError, even with a permissive isRetryableError', async () => {
+    const abortError = new DOMException('The operation was aborted', 'AbortError');
+    const fetchMock = vi.fn().mockRejectedValue(abortError);
+    vi.stubGlobal('fetch', fetchMock);
+
+    await expect(
+      fetchWithRetry('http://abortedRequest', {
+        isRetryableError: () => true
+      })
+    ).rejects.toThrow(
+      'Network error fetching http://abortedRequest: The operation was aborted'
+    );
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+  });
 });
 
 describe('fetchWithRetryRaw', () => {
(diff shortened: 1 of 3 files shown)


packages/c2pa-utilities/src/fetchWithRetry.ts
```diff
@@ -85,6 +85,28 @@ export interface FetchWithRetryOptions {
    * always retried regardless of this predicate.
```
tmathern: Isn't `isRetryableError` involved in this decision too now?

packages/c2pa-utilities/src/fetchWithRetry.spec.ts
```diff
@@ -368,6 +368,33 @@ describe('fetchWithRetry', () => {
     const result = await fetchWithRetry('http://retryAfterOn503');
     expect(result).toBe('resolved after 503 retry-after');
   });
+
+  test('honors a custom isRetryableError', async () => {
```
tmathern, line 372:
```
export async function fetchWithRetryRaw(
  input: string,
  init?: RequestInit,
  options?: FetchWithRetryOptions
)
```

Maybe also add a case where `RequestInit` is set?

packages/c2pa-utilities/src/fetchWithRetry.ts
```diff
@@ -167,16 +189,18 @@ export async function fetchWithRetryRaw(
...
     calculateBackoffMs(attempt, initialRetryDelayMs, maxRetryDelayMs);
 
   for (let attempt = 0; ; attempt++) {
     let res: Response;
     try {
-      res = await fetch(input, init);
+      res = await fetchFn(input, init);
     } catch (e) {
-      if (attempt < maxRetries) {
+      if (!isAbortError(e) && attempt < maxRetries && isRetryableError(e)) {
```
tmathern:
```suggestion
      if (!isAbortError(e) && isRetryableError(e) && attempt < maxRetries) {
```
Debatable: I would test if we can retry before we even check the attempts (again, debatable/optional).

packages/c2pa-utilities/src/fetchWithRetry.ts
```diff
@@ -85,6 +85,28 @@ export interface FetchWithRetryOptions {
...
+   * (like `wretch` middleware's `next`, or a test mock).
+   */
+  fetch?: (input: string, init?: RequestInit) => Promise<Response>;
+}
+
+/**
+ * @param error The error to check, usually from a `fetch` call.
+ * @returns Whether `error` represents a deliberate `AbortError`.
+ */
+function isAbortError(error: unknown): boolean {
```
tmathern, line 109: I don't think this `Abort` is related to AbortController?
Node: https://nodejs.org/api/globals.html#class-abortcontroller
Web: https://developer.mozilla.org/en-US/docs/Web/API/AbortController

We should consider adding AbortControllers in there, at least potentially offer a caller to propagate them into those calls (probably not this PR, but could be a nice follow-up).
Note this would change the checks here for the error. And the tiemout logic as well (since you'd check the signal).

tmathern (approved):
LGTM, comments are optional notes.

packages/c2pa-utilities/src/fetchWithRetry.ts
ale-adobe: Updated comments between here and below for consistency and clarity.

packages/c2pa-utilities/src/fetchWithRetry.ts
ale-adobe: Agreed, this makes more sense.

packages/c2pa-utilities/src/fetchWithRetry.ts
ale-adobe: An abort triggered by an `AbortController` would be classified as an abort error and would return `true` from this function.

packages/c2pa-utilities/src/fetchWithRetry.spec.ts
ale-adobe: Good idea; added.

PR #191, opened 2026-08-14
ci: Make GitHub Action `release.yml` more robust and idempotent

ale-adobe:
Make GitHub Action `release.yml` more robust and idempotent:

- Use `changeset publish` to publish packages, which intelligently skips over packages that have already been published.
- Add retry with 10s backoff when grabbing release ID to account for cases where the ID is delayed due to eventual consistency.
- Overwrite existing assets in case a previous attempt to publish failed partway, rather than failing because it sees that something was already "published".

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -76,6 +76,15 @@ jobs:
             echo "::error::Could not find @contentauth/c2pa-node in publishedPackages" >&2
             exit 1
           fi
+
+          # publishedPackages reflects what changesets/action *attempted*, not
+          # what actually landed on the registry, so verify directly against npm
+          # before releasing artifacts.
+          if ! npm view "@contentauth/c2pa-node@${version}" version >/dev/null 2>&1; then
+            echo "::error::@contentauth/c2pa-node@${version} is not on the npm registry; refusing to build/upload binaries for it" >&2
+            exit 1
+          fi
+
           echo "PUBLISHED_VERSION=$version" >> "$GITHUB_ENV"
 
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -76,6 +76,15 @@ jobs:
...
             exit 1
           fi
+
+          # publishedPackages reflects what changesets/action *attempted*, not
+          # what actually landed on the registry, so verify directly against npm
+          # before releasing artifacts.
+          if ! npm view "@contentauth/c2pa-node@${version}" version >/dev/null 2>&1; then
+            echo "::error::@contentauth/c2pa-node@${version} is not on the npm registry; refusing to build/upload binaries for it" >&2
+            exit 1
+          fi
```
ale-adobe, line 86: GH actions was previously reporting that a package was published because it attempted to do so. It would have a release tag and partially complete release on the GitHub page, despite not actually succeeding in publishing the package to `npmjs`.

.github/workflows/release.yml
```diff
@@ -159,28 +168,66 @@ jobs:
...
+              echo "id=$id" >> "$GITHUB_OUTPUT"
+              exit 0
+            fi
+
+            echo "GitHub release for tag ${tag} not found yet (attempt ${attempt}/5, HTTP ${http_code}); retrying in 10s..."
+            sleep 10
+          done
+
+          echo "::error::Could not find a GitHub release tagged ${tag} after retrying" >&2
+          exit 1
```
ale-adobe, line 193: Extra resiliency.

tmathern (approved):

PR #201, opened 2026-08-27
feat(c2pa-utilities): Introduce Context class

ale-adobe:
## Summary
This PR introduces the `Context` class, mirroring the one used in native SDKs, as a wrapper around `Settings` for configuring `Reader` and `Builder` objects. This class is adopted by `c2pa-web` in https://github.com/contentauth/c2pa-js/pull/202 and by `c2pa-node` in https://github.com/contentauth/c2pa-js/pull/204 as the new way to configure SDK behavior.

The changes in this PR are considered **breaking changes**.

## Changes
- Added a `Context` class wrapping `Settings`, mirroring `Context` used in native SDKs.
  - `Context` is immutable and is meant to supplied when creating a `Reader` or `Builder`, customizing that instance's behavior.
- Simplified `resolveSettings` by no longer accepting overrides, since the concept of providing per-call overrides no longer applies. Users should create and merge their settings as needed, then pass them into a `Context` that goes into the `Reader` or `Builder` that they wish to configure.
  - Added temporary shims in `c2pa-web` to accommodate the signature change; will be removed in https://github.com/contentauth/c2pa-js/pull/202.
- Added `withDefaultSettings()`, a synchronous counterpart to `resolveSettings` that applies this package's defaults without the async trust-anchor URL resolution, for use in `c2pa-node` which needs the defaults without supporting that fetch (yet).

### DIFF
--- a/.changeset/blue-jobs-allow.md
+++ b/.changeset/blue-jobs-allow.md
@@ -1,5 +1,5 @@
 ---
-'@contentauth/c2pa-utilities': minor
+'@contentauth/c2pa-utilities': major
 '@contentauth/c2pa-web': patch
 ---
 
--- a/.changeset/blue-jobs-allow.md
+++ b/.changeset/blue-jobs-allow.md
@@ -1,0 +1,5 @@
+---
+'@contentauth/c2pa-utilities': minor
+---
+
+Introduce Context, deprecate thread-local Settings.
--- a/.changeset/blue-jobs-allow.md
+++ b/.changeset/blue-jobs-allow.md
@@ -1,6 +1,5 @@
 ---
 '@contentauth/c2pa-utilities': minor
-'@contentauth/c2pa-wasm': patch
 '@contentauth/c2pa-web': patch
 ---
 
--- a/.changeset/blue-jobs-allow.md
+++ b/.changeset/blue-jobs-allow.md
@@ -1,5 +1,7 @@
 ---
 '@contentauth/c2pa-utilities': minor
+'@contentauth/c2pa-wasm': patch
+'@contentauth/c2pa-web': patch
 ---
 
 Introduce Context, deprecate thread-local Settings.
--- a/.changeset/blue-jobs-allow.md
+++ b/.changeset/blue-jobs-allow.md
@@ -3,4 +3,4 @@
 '@contentauth/c2pa-web': patch
 ---
 
-Introduce Context, deprecate thread-local Settings.
+Introduce Context class as a wrapper around Settings for configuring Readers and Builders. Add temporary shim in c2pa-web.
--- a/.changeset/blue-jobs-allow.md
+++ b/.changeset/blue-jobs-allow.md
@@ -1,5 +1,5 @@
 ---
-'@contentauth/c2pa-utilities': major
+'@contentauth/c2pa-utilities': minor
 '@contentauth/c2pa-web': patch
 ---
 
--- a/packages/c2pa-utilities/README.md
+++ b/packages/c2pa-utilities/README.md
@@ -14,9 +14,53 @@ Complete API documentation is generated from TypeScript source using [TypeDoc](h
 
 ## Utilities
 
-### Settings
+### `Context` and `Settings`
 
-Helpers for building, merging, and serializing the `Settings` object consumed by `c2pa-web`'s and `c2pa-node`'s `Reader`/`Builder` constructors.
+`Settings` is a plain, JSON-serializable object describing SDK behavior — trust anchors, verification options, and builder options. `Context` is a small, immutable wrapper around `Settings`, and is the recommended way to configure the SDK. Clients should create and attach a `Context` once at SDK-initialization time, which then configures every `Reader`/`Builder` created afterwards. It is not passed to individual `Reader`/`Builder` instances or their per-call methods. See [`c2pa-web`'s README](../c2pa-web/README.md#configuring-the-sdk-with-context) for an example.
+
+#### Creating a `Context`
+
+```typescript
+import { Context } from '@contentauth/c2pa-utilities';
+
+const context = new Context({
+  verify: {
+    verifyTrust: true
+  },
+  trust: {
+    trustAnchors: 'https://example.com/trust-anchors.pem'
+  }
+});
+```
+
+#### Combining `Settings`
+
+`Context` doesn't merge settings for you — each `Context` just holds whatever single `Settings` object was passed to its constructor. To combine more than one `Settings` source, merge them first with `mergeSettings()`, then construct a `Context` from the single, merged result:
+
+```typescript
+import { Context, mergeSettings } from '@contentauth/c2pa-utilities';
+
+const base = { verify: { verifyTrust: true } };
+const override = { verify: { verifyAfterSign: true } };
+
+const context = new Context(mergeSettings(base, override));
+// context.settings is { verify: { verifyTrust: true, verifyAfterSign: true } }
+```
+
+#### Resolving a `Context` to JSON
+
+Bindings pass settings across their JS-to-wasm/native boundary as a JSON string. `toJson()` resolves any trust-anchor URLs embedded in the settings (fetching and validating them) and serializes the result:
+
+```typescript
+const contextJson = await context.toJson();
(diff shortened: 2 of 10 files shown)


packages/c2pa-utilities/src/settings.ts
```diff
@@ -34,7 +34,7 @@ import { snakeCaseify, type SettingsObjectType } from './caseConversion.js';
  *   }
  * };
  *
- * const settingsJson = await resolveSettings(settings, undefined);
```
tmathern, line 37: https://github.com/contentauth/c2pa-js/blob/main/packages/c2pa-web/src/lib/c2pa.ts#L73

This also has a second undefined param, why not change it there too?

packages/c2pa-utilities/src/context.ts
```diff
@@ -0,0 +1,53 @@
...
+   * of this Context's settings should only ever happen once. If different fetch-with-retry options
+   * need to be provided, callers should create a separate `Context` and then call `toJson()`
+   * with the desired options.
+   *
+   * @param options Optional configurations for fetch-with-retry, used when resolving trust-anchor
+   * URLs. Only consulted on the first call.
+   * @returns A JSON-serialized string of the resolved settings.
+   */
+  toJson(options?: FetchWithRetryOptions): Promise<string> {
+    this._jsonPromise ??= resolveSettings(this.settings, options);
```
tmathern: Does this cache only on success? Or would it cache failures if resolving settings fail? Asking because we should be able to retry on failure.

packages/c2pa-utilities/src/settings.ts
```diff
@@ -150,40 +150,49 @@ export interface BuilderSettings {
 // Defaults
 // =================================
 
-const DEFAULT_SETTINGS: Settings = {
+/**
+ * This package's default settings, applied by {@link withDefaultSettings} and
+ * {@link resolveSettings} so a `Reader`/`Builder` gets sane behavior even with no settings
+ * provided at all.
+ */
+export const DEFAULT_SETTINGS: Settings = {
```
tmathern: Can we make those defaults read-only explicitly? if someone wants to use them as base, they should clone and modify their own version.

.changeset/blue-jobs-allow.md
```diff
@@ -0,0 +1,6 @@
+---
+'@contentauth/c2pa-utilities': minor
```
tmathern: I think actually the resolveSettings change may turn this into a breaking change?

packages/c2pa-utilities/src/context.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    // Result should contain the defaults that the given settings are merged with.
+    expect(result).toEqual(
+      JSON.stringify({
+        builder: { generate_c2pa_archive: true },
+        verify: { verify_trust: false }
+      })
+    );
+  });
+
+  test('toJson() memoizes: repeated calls return the exact same Promise', async () => {
```
tmathern: Is the promise memoized, or its resolution?

packages/c2pa-utilities/src/context.ts
```diff
@@ -0,0 +1,53 @@
...
+ * NOTICE: Adobe permits you to use, modify, and distribute this file in
+ * accordance with the terms of the Adobe license agreement accompanying
+ * it.
+ */
+
+import type { FetchWithRetryOptions } from './fetchWithRetry.js';
+import { resolveSettings, type Settings } from './settings.js';
+
+/**
+ * A `Context` configures the behavior of a single `Reader`/`Builder`.
```
tmathern:
```suggestion
 * A `Context` configures the behavior of a `Reader`/`Builder`. 
```
It is a bit unclear just from the comment, but a Context could be reused by multiple object.

packages/c2pa-utilities/src/settings.spec.ts
```diff
@@ -244,7 +184,7 @@ describe('settings', () => {
           )
         );
 
-        const result = await resolveSettings(undefined, {
```
tmathern, line 247: This does look like a breaking change...

packages/c2pa-utilities/README.md
```diff
@@ -14,9 +14,53 @@ Complete API documentation is generated from TypeScript source using [TypeDoc](h
 
 ## Utilities
 
-### Settings
+### `Context` and `Settings`
 
-Helpers for building, merging, and serializing the `Settings` object consumed by `c2pa-web`'s and `c2pa-node`'s `Reader`/`Builder` constructors.
+`Settings` is a plain, JSON-serializable object configuring SDK behavior around trust anchors, verification options, and `Reader`/`Builder` options. `Context` is a small, immutable wrapper around `Settings`, and is the recommended way to configure a `Reader`/`Builder`. `Context` objects are passed directly to the call that creates a `Reader`/`Builder` instance, so one running SDK instance can freely create many different `Reader`/`Builder`s, each with its own `Context`. See [`c2pa-web`'s README](../c2pa-web/README.md#configuring-behavior-with-context) for an example.
```
tmathern: Maybe clarify that if a context changes after a builder/reader is created with that context, the context change does not propagate. It is frozen in time at creation of reader/builder (as far as the native side goes).

packages/c2pa-web/src/lib/builder.ts
```diff
@@ -304,7 +304,11 @@ export function createBuilderFactory(worker: WorkerManager, settings?: Settings)
 
   return {
     async new(settings?: Settings) {
-      const settingsJson = await resolveSettings(baseSettings, settings);
+      // TODO: temporary shim for c2pa-utilities' resolveSettings signature change
+      // (single-argument now); removed once c2pa-web adopts Context in a follow-up PR.
```
tmathern, line 308: (Repeating it, I know: but that means eventually this is a breaking change).

tmathern (commented):
See inline comments.

packages/c2pa-utilities/src/context.ts
ale-adobe: This memo-izes the promise itself.

.changeset/blue-jobs-allow.md
ale-adobe: Yes, it is, which is why it's listed as minor rather than patch. We've been using minor patch bumps for breaking changes, since we haven't "gone to 1.0" yet.

I agree this is starting to get a bit cumbersome though, so I'm open to making this be the first major patch and have this Context work kick off 1.0 for these packages.

packages/c2pa-utilities/src/context.spec.ts
ale-adobe: The promise is memoized.

packages/c2pa-utilities/src/context.ts
ale-adobe: Good point. I'll make it explicit that it can be re-used by multiple objects as an intended use case.

packages/c2pa-utilities/src/settings.spec.ts
ale-adobe: It is. 😄

packages/c2pa-web/src/lib/builder.ts
ale-adobe: Yes; breaking changes all around!

packages/c2pa-utilities/src/settings.ts
ale-adobe: Thanks; it actually had two, and I missed that when cleaning it up. This is currently passing in `undefined` for the fetch-with-retry options, which doesn't do anything, so I'll remove it there too.

packages/c2pa-utilities/src/context.ts
ale-adobe: Updating to not cache failures, which would make the context un-useable afterwards. This way, we can retry on future calls to `toJson()`.

cdmurph32 (approved):
LGTM

.changeset/blue-jobs-allow.md
ale-adobe: As discussed in Slack, we'll be sticking with minor version bumps.

tmathern (approved):

PR #202, opened 2026-08-27
feat(c2pa-web): Use Context to configure Readers/Builders

ale-adobe:
## Summary
This PR brings in the `Context` class introduced in `c2pa-utilities` (see https://github.com/contentauth/c2pa-js/pull/201) in `c2pa-web`, replacing the SDK-wide `Settings` object and the `Reader`/`Builder` factory pattern with static constructors and per-instance `Context` objects. This mirrors how `c2pa-node` works (see https://github.com/contentauth/c2pa-js/pull/204).

In addition, we remove the dead thread-local `Settings` path using `loadSettings` from `c2pa-wasm`, since `Settings` are now passed explicitly via the `Context` object.

### `C2pa` class (`c2pa.ts`)
- Renamed `C2paSdk` to just `C2pa`.
- Simplified `C2pa` class: now functions as a way to start up a web worker instance and load the Wasm binary. The returned handle from `createC2pa()` is passed in directly when creating a `Reader` or `Builder`.

### `Reader` (`reader.ts`) and `Builder` (`builder.ts`)
- Deprecated the existing `ReaderFactory` / `createReaderFactory` and `BuilderFactory` / `createBuilderFactory` patterns, but still retain them for now in order to not introduce breaking changes just yet.
- Added static methods to `Reader` and `Builder` for constructing new instances, which accept the `C2pa` handle above, plus an optional `Context` to configure their respective behaviors. This is the new, recommended way to create them.

New example usage:
```typescript
import { createC2pa, Reader, Builder, Context } from '@contentauth/c2pa-web';
import wasmSrc from '@contentauth/c2pa-web/resources/c2pa.wasm?url';

const c2pa = await createC2pa({ wasmSrc });

const context = new Context({
  verify: {
    verifyTrust: true
  },
  trust: {
    trustAnchors: 'https://example.com/trust-anchors.pem'
  }
});

const reader = await Reader.fromBlob(c2pa, blob.type, blob, context);
const builder = await Builder.new(c2pa, context);
```

### Thread-local `Settings`
- Removed `c2pa-wasm`'s dead thread-local settings path (`loadSettings`/`Settings::from_string`). The one caller (`c2pa-web`'s worker init) is removed along with it.


packages/c2pa-web/src/lib/reader.spec.ts
```diff
@@ -275,47 +256,9 @@ describe('reader', () => {
         });
 
         await expect(
-          c2pa.reader.fromBlobFragment(initBlob.type, initBlob, fragmentBlob)
+          Reader.fromBlobFragment(c2pa, initBlob.type, initBlob, fragmentBlob)
         ).rejects.toThrow(AssetTooLargeError);
       });
-
-      test('should inherit global settings when per-call settings are provided', async () => {
```
ale-adobe, line 282: No more support for per-call settings.

packages/c2pa-web/src/lib/reader.spec.ts
```diff
@@ -105,97 +98,82 @@ describe('reader', () => {
...
 
         const manifestStore = await reader!.manifestStore();
 
-        // Using the overrideSettings, the asset is trusted.
         expect(manifestStore).toEqual(C_with_CAWG_data_trusted_ManifestStore);
 
         c2pa.dispose();
       });
 
-      test('should inherit global settings when per-call settings are provided', async () => {
```
ale-adobe, line 128: No more support for per-call settings.

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -35,22 +35,14 @@ describe('builder', () => {
         });
       });
 
-      test('should use local "context" settings when provided', async () => {
```
ale-adobe, line 38: No more support for per-call settings.

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -66,97 +58,46 @@ describe('builder', () => {
         expect(ingredientFailureCodes).toContain('signingCredential.untrusted');
       });
 
-      test('should inherit global settings when per-call settings are provided', async () => {
```
ale-adobe, line 69: No more support for per-call settings.

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -66,97 +58,46 @@ describe('builder', () => {
...
+          definitionA.ingredients?.[0].validation_results?.activeManifest?.failure.map(
             (entry) => entry.code
-          );
-
-        expect(ingredientFailureCodes).toContain('signingCredential.untrusted');
-
-        c2pa.dispose();
-      });
-
-      test('should allow per-call settings to override conflicting global settings', async () => {
```
ale-adobe, line 106: No more support for per-call settings.

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -169,44 +110,7 @@ describe('builder', () => {
             (entry) => entry.code
           );
 
-        // Global settings are configured to verify trust, so the result should be untrusted.
         expect(ingredientFailureCodes).toContain('signingCredential.untrusted');
-
-        c2pa.dispose();
-      });
-
-      test('should allow per-call settings to override conflicting global settings', async () => {
```
ale-adobe, line 178: No more support for per-call settings.

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -269,43 +165,6 @@ describe('builder', () => {
           );
 
         expect(ingredientFailureCodes).toContain('signingCredential.untrusted');
-
-        c2pa.dispose();
-      });
-
-      test('should allow per-call settings to override conflicting global settings', async () => {
```
ale-adobe, line 276: No more support for per-call settings.

packages/c2pa-web/src/lib/builder.ts
```diff
@@ -195,335 +371,157 @@ export interface Builder {
...
+   * from the top-level settings passed to {@link createC2pa}.
+   * @deprecated Use {@link Builder.fromArchive} with a `Context` instead.
+   */
+  fromArchive: (archive: Blob, settings?: Settings) => Promise<Builder>;
 }
 
 /**
- * @param worker - Worker (via WorkerManager) to be associated with this reader factory.
- * @param settings - Optional settings to be used for all builders.
+ * @deprecated Use `Builder.new`/`Builder.fromDefinition`/`Builder.fromArchive` with a `Context`
```
ale-adobe, line 498: Existing `BuilderFactory` is still here, but now marked as deprecated.

packages/c2pa-web/src/lib/reader.ts
```diff
@@ -101,103 +177,82 @@ export interface Reader {
...
+  fromBlobFragment: (
+    format: string,
+    init: Blob,
+    fragment: Blob,
+    settings?: Settings
+  ) => Promise<Reader | null>;
+}
 
+/**
+ * @deprecated Use `Reader.fromBlob`/`Reader.fromBlobFragment` with a `Context` instead.
```
ale-adobe, line 231: Existing `ReaderFactory` is still here, but now marked as deprecated.

packages/c2pa-wasm/src/settings.rs
```diff
@@ -1,26 +0,0 @@
...
- * See: https://github.com/wasm-bindgen/wasm-bindgen/issues/4961
- */
-
-/// Accepts a JSON-serialized string to be loaded as c2pa-rs settings.
-#[wasm_bindgen(js_name = loadSettings)]
-pub fn load_settings(settings: &str) -> Result<(), JsString> {
-    c2pa::settings::Settings::from_string(settings, "json").map_err(WasmError::other)?;
-
-    Ok(())
-}
```
ale-adobe, line 26: Gone, since we now pass settings through `Context`.

packages/c2pa-web/src/lib/builder.ts
```diff
@@ -7,7 +7,7 @@
  * it.
```
ale-adobe, line 7: The diff in this file is a bit scary, but not much has actually changed at a high level.

- Helper functions are pulled out to the top-level as free functions.
- Functions that used to be in the `createBuilder` factory are now static functions on the `Builder` class.
- The old `BuilderFactory` is maintained at the bottom, but marked deprecated.
- *New*: Functions for constructing a new `Builder` accept the `C2pa` object and the optional `Context`.

packages/c2pa-web/src/lib/reader.ts
```diff
@@ -10,88 +10,164 @@
 import { Manifest, ManifestStore } from '@contentauth/c2pa-types';
```
ale-adobe, line 10: The diff in this file is a bit scary, but not much has actually changed at a high level.

- Helper functions are pulled out to the top-level as free functions.
- Functions that used to be in the `createReader` factory are now static functions on the `Reader` class.
- The old `ReaderFactory` is maintained at the bottom, but marked deprecated.
- *New*: Functions for constructing a new `Reader` accept the `C2pa` object and the optional `Context`.

packages/c2pa-web/src/lib/c2pa.ts
```diff
@@ -29,55 +29,72 @@ export interface Config {
...
-  const settingsString = await resolveSettings(settings, undefined);
-  const worker = await createWorkerManager({ wasm, workerSrc, settingsString });
+  const worker = await createWorkerManager({ wasm, workerSrc });
 
-  return {
-    reader: createReaderFactory(worker, settings),
-    builder: createBuilderFactory(worker, settings),
+  const c2pa: C2pa = {
+    worker,
+    reader: undefined as unknown as ReaderFactory,
```
ale-adobe, line 90: A workaround to satisfy the shape of the object before creating the factories below, since we are still trying to keep around the factories as deprecated paths to avoid breaking changes.

packages/c2pa-web/src/common.ts
```diff
@@ -7,15 +7,11 @@
  * it.
  */
 
-export type * from './lib/c2pa.js';
+export type { Config, C2pa } from './lib/c2pa.js';
```
tmathern, line 10: export changes may make this a breaking change if exposed types are different

packages/c2pa-web/src/lib/c2pa.ts
```diff
@@ -29,55 +29,72 @@ export interface Config {
...
 
   /**
+   * @deprecated Use a `Context` to construct a `Reader`/`Builder` via their static methods instead.
    * Settings for the SDK. Will be inherited by created builders and readers.
    */
   settings?: Settings;
 }
 
-export interface C2paSdk {
+export interface C2pa {
```
tmathern, line 38: The rename may make this a breaking change (another reason)

packages/c2pa-web/src/lib/worker/rpc.ts
```diff
@@ -15,7 +15,7 @@ import { channel } from 'highgain';
 
 // Define browser-to-worker RPC interface
 const { createTx, rx } = channel<{
-  initWorker: (module: WebAssembly.Module, settings?: string) => void;
+  initWorker: (module: WebAssembly.Module) => void;
```
tmathern, line 18: Breaking change?

packages/c2pa-web/src/lib/builder.ts
tmathern: Can we summarize the prose a little bit maybe, to make it less scary?

packages/c2pa-web/src/lib/builder.ts
```diff
@@ -17,101 +17,207 @@ import type {
...
    * Create a {@link Builder} with a minimal manifest definition as its initial state.
-   * @param settings Optional context settings for the builder. Will override any values inherited by the top-level settings passed to createC2pa.
+   *
+   * @param c2pa The `C2pa` instance (from {@link createC2pa}) to create this builder on.
+   * @param context Optional `Context` configuring this builder's behavior.
    * @returns A {@link Builder} object.
    */
-  new: (settings?: Settings) => Promise<Builder>;
+  static async new(c2pa: C2pa, context: Context = new Context()): Promise<Builder> {
+    const settingsJson = await context.toJson();
```
tmathern, line 87: Can a context be returned as null if there was an error in `context.toJson`?

packages/c2pa-web/src/lib/reader.ts
```diff
@@ -10,88 +10,164 @@
...
-  fromBlob: (
+  static async fromBlob(
+    c2pa: C2pa,
     format: string,
     blob: Blob,
-    settings?: Settings
-  ) => Promise<Reader | null>;
+    context: Context = new Context()
+  ): Promise<Reader | null> {
+    if (!isSupportedReaderFormat(format)) {
```
tmathern, line 64: Is this still needed, now that the Rust SDK attempts auto-detect?

tmathern (dismissed):
Double-check with the native SDK, but I think you can simplify the Reader format check.

packages/c2pa-web/src/common.ts
ale-adobe: Yes, correct, alongside the rename to just `C2pa`.

packages/c2pa-web/src/lib/builder.ts
ale-adobe: Not anymore, since I made some changes in https://github.com/contentauth/c2pa-js/pull/201. It will either resolve into a `string` or throw if something goes wrong. If an error is thrown, that error will propagate through the constructor here.

packages/c2pa-web/src/lib/builder.ts
ale-adobe: I can try to trim some of the docs and comments, but most of that was existing from previous AI compliance work that Colin did.

packages/c2pa-web/src/lib/worker/rpc.ts
ale-adobe: Yes. Changesets are marked as minor for the breaking change, but willing to make this the major version bump to go to 1.0.

packages/c2pa-web/src/lib/reader.ts
ale-adobe: Maybe not. But I think that's out-of-scope for this PR, so let me revisit that separately. 

(I'm inclined to believe that you're right, since this is something we flagged in my proposal originally. Node doesn't do this, and Rust is doing the heavy lifting there as far as I can tell. So I don't see why Web shouldn't do the same thing.)

Regardless, filed https://jira.corp.adobe.com/browse/CAI-13611 this in the epic for me to follow up on later.

cdmurph32 (approved):
LGTM

packages/c2pa-web/src/lib/reader.spec.ts
```diff
@@ -21,8 +23,6 @@ import C_with_CAWG_data_ManifestStore from 'test/manifests/C_with_CAWG_data.js';
 
 import C_with_CAWG_data_trusted_ManifestStore from 'test/manifests/C_with_CAWG_data_trusted.js';
 import C_with_CAWG_data_untrusted_ManifestStore from 'test/manifests/C_with_CAWG_data_untrusted.js';
-import C_with_CAWG_data_cawg_only_trusted_ManifestStore from 'test/manifests/C_with_CAWG_data_cawg_only_trusted.js';
-import C_with_CAWG_data_signature_only_trusted_ManifestStore from 'test/manifests/C_with_CAWG_data_signature_only_trusted.js';
```
ale-adobe, line 25: No longer needed; also deleted the corresponding test fixtures since they are no longer used.

tmathern (approved):

PR #204, opened 2026-08-27
feat(c2pa-node): Use Context to configure Readers/Builders

ale-adobe:
## Summary
This PR brings in the `Context` class introduced in `c2pa-utilities` (see https://github.com/contentauth/c2pa-js/pull/201) in `c2pa-web`, replacing the SDK-wide `Settings` object and the `Reader`/`Builder` factory pattern with static constructors and per-instance `Context` objects. This mirrors how `c2pa-web` works (see https://github.com/contentauth/c2pa-js/pull/202).

This PR is fully backward-compatible.

## `Reader` (`Reader.ts`)
- `Reader` creation functions now accept a `Context` in addition to the deprecated raw `C2paSettings` string/object

## `Builder` (`Builder.ts`)
- `Builder.new`/`Builder.withJson` stay synchronous and deprecated, for backward compatibility, and will be removed in a future version.
- New `Builder.newAsync`/`Builder.withJsonAsync` are the async replacements that accept the new `Context`
  - Function signatures are declared async for forward-compatibility. Future work will involve JS-side trust-anchor resolution to mirror what happens over in `c2pa-web` and maintain behavioral parity.
- New `resolveSettingsForNeon` internal helper added to `Settings.ts` to help resolve settings as it currently does without going through trust-anchor resolution (yet).

### DIFF
--- a/packages/c2pa-node/js-src/Builder.ts
+++ b/packages/c2pa-node/js-src/Builder.ts
@@ -74,15 +74,8 @@ export class Builder implements BuilderInterface {
    * Create a Builder with a minimal manifest definition, configured by an optional `Context`.
    * Prefer this over the deprecated {@link Builder.new}.
    *
-   * ## Why `async`, when nothing here awaits anything yet
-   *
-   * `resolveSettingsForNeon` doesn't resolve trust-anchor URLs today (see its doc comment in
-   * `Settings.ts` for the proposal to add that), so building a `Context`-backed Builder is
-   * currently just as synchronous as {@link Builder.new}. This method is `async` anyway so that
-   * adding that trust-anchor resolution later — the natural next step once it's signed off —
-   * only has to change what happens *inside* this method, not its public signature. Landing the
-   * `Promise<Builder>` signature now, ahead of the work that actually needs it, means callers
-   * that adopt `Context` today never have to migrate a second time when that work lands.
+   * Declared async for forward compatibility. Upcoming work will allow c2pa-node
+   * to resolve trust anchors at construction time.
    *
    * @param context A `Context` object containing configuration settings for this Builder.
    */
@@ -94,7 +87,7 @@ export class Builder implements BuilderInterface {
 
   /**
    * @deprecated Use {@link Builder.withJsonAsync} instead, passing a `Context`. Will be removed
-   * in a future major version.
+   * in a future version.
    * @param settings A raw `C2paSettings` string/object.
    */
   static withJson(json: Manifest, settings?: C2paSettings): Builder {
@@ -111,8 +104,8 @@ export class Builder implements BuilderInterface {
    * Create a Builder from a manifest definition, configured by an optional `Context`. Prefer
    * this over the deprecated {@link Builder.withJson}.
    *
-   * `async` for the same forward-compatibility reason as {@link Builder.newAsync} — see its doc
-   * comment.
+   * Declared async for forward compatibility. Upcoming work will allow c2pa-node
+   * to resolve trust anchors at construction time.
    */
   static async withJsonAsync(
     json: Manifest,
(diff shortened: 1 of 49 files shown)


packages/c2pa-node/js-src/types.d.ts
```diff
@@ -180,7 +180,7 @@ export interface HashedUri {
 
 /**
  * Optional settings that can be provided when creating a Reader or Builder.
- * Can be a JSON string, TOML string, or a settings object.
+ * Can be a JSON string or a settings object.
```
ale-adobe, line 183: Removing mention of TOML since we only really work with JSON now, AFAIK.

packages/c2pa-node/js-src/Settings.ts
```diff
@@ -29,3 +31,27 @@ export async function loadSettingsFromFile(filePath: string): Promise<string> {
...
+ * or a `Context`) into the JSON string the native library expects.
+ * 
+ * Unlike `c2pa-web`'s `Context.toJson()`, this does not resolve trust-anchor URLs: `c2pa-node`
+ * doesn't perform trust-anchor URL fetching yet, so a `Context`'s settings are only merged with
+ * this package's defaults and serialized here.
+ */
+export function resolveSettingsForNeon(
+  settingsOrContext: C2paSettings | Context | undefined,
+): string | undefined {
+  if (settingsOrContext === undefined) {
```
tmathern: I would verify there is no side-effects with strings/values like `null` or `undefined`, since there is likely to be serialization ebtween the Reader/Builder call and this somewhere.

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -38,39 +40,79 @@ import type {
...
+   *
+   * Declared async for forward compatibility. Upcoming work will allow c2pa-node
+   * to resolve trust anchors at construction time.
+   */
+  static async withJsonAsync(
+    json: Manifest,
+    context?: Context,
+  ): Promise<Builder> {
+    const jsonString = stringifyManifestDefinition(json);
+    const settingsStr = resolveSettingsForNeon(context);
```
tmathern, line 120: It is `context?: Context`, so can it be null here actually? Should it be allowed to be null/undefined, because then it falls probably back to no context?

packages/c2pa-node/js-src/Builder.ts
```diff
@@ -143,15 +185,15 @@ export class Builder implements BuilderInterface {
...
     asset: SourceAsset,
-    settings?: C2paSettings,
+    settingsOrContext?: C2paSettings | Context,
   ): Promise<Builder> {
-    const settingsStr = settings
-      ? typeof settings === "string"
-        ? settings
-        : JSON.stringify(settings)
-      : undefined;
+    const settingsStr = resolveSettingsForNeon(settingsOrContext);
```
tmathern, line 202: `settingsOrContext?: C2paSettings | Context,`, so is settingsOrContext optional and potentially null/undefined at this point?

tmathern (commented):
If a context is meant to hold all settings, including default settings (e.g. trust), we should apply extra care making sure it is set when we really, really need it to have its settings. I know there is no easy overload in js/ts, but making the context optional may be dangerous too, as it would be easy to forget about/missable to use it.

packages/c2pa-node/js-src/Builder.ts
ale-adobe: Good catch -- updated so that null and undefined are treated the same way (no context provided -> use defaults).

packages/c2pa-node/js-src/Builder.ts
ale-adobe: As above, updating so that we treat null and undefined the same way.

cdmurph32 (approved):
LGTM

tmathern (approved):

PR #208, opened 2026-09-01
fix(c2pa-node): Switch `fs-extra` to default import from namespace import

ale-adobe:
This PR switches the `fs-extra` import in `c2pa-node` to use a default import pattern rather than a namespace import, which makes it so that it gets the module's actual `module.exports` object instead of a synthetic namespace built from statically-detected named exports. This gap breaks calls like `fs.stat()` in `c2pa-node` at actual runtime, and was not caught by Vitest's testing envrionment.


cdmurph32 (approved):
Approved

PR #213, opened 2026-09-08
feat(c2pa-web): Remove supported format checks for Reader

ale-adobe:
This PR removes supported format checks from `c2pa-web`'s `Reader` class. This logic is already implemented by the native library, so there's no need to duplicate it here and also potentially introduce behavior drift. `c2pa-node` already relies on the native library for these checks, so this PR brings `c2pa-web` up to parity.


scouten-adobe (approved):

cdmurph32 (approved):
LGTM

PR #214, opened 2026-09-09
fix(c2pa-web): Incomplete sanitization of manifest store labels

ale-adobe:
This PR follows up on #145's fix, which was incomplete. 

In this PR:
- The sanitized manifest store now avoids clobbering the active manifest when it finds the bad `__proto__` label by re-assigning it in-place, rather than assigning the sanitized manifest over the active one.
- Add unit test to verify the above case.


cdmurph32 (approved):
LGTM

PR #220, opened 2026-09-15
fix: retry npm registry check for `c2pa-node` before building binaries

ale-adobe:
Follow-up PR to https://github.com/contentauth/c2pa-js/pull/191.

The "Resolve published c2pa-node version" step in `release.yml` checked npm just once via `npm view` before deciding whether to build and upload native binaries. When the registry hadn't yet propagated a just-published version, every platform leg of the `release-node-binaries` job failed with "not on the npm registry", leaving the GitHub release for that version with no binary assets.

Add the same retry/backoff already used a few steps later for the release-id lookup, which has the identical eventual-consistency rationale, making the overall job more robust to race conditions.


.github/workflows/release.yml
```diff
@@ -79,9 +79,21 @@ jobs:
...
           # publishedPackages reflects what changesets/action *attempted*, not
           # what actually landed on the registry, so verify directly against npm
-          # before releasing artifacts.
-          if ! npm view "@contentauth/c2pa-node@${version}" version >/dev/null 2>&1; then
-            echo "::error::@contentauth/c2pa-node@${version} is not on the npm registry; refusing to build/upload binaries for it" >&2
+          # before releasing artifacts. The registry can take a few seconds to
+          # serve a just-published version, so retry rather than failing on
+          # the first miss (same reasoning as the release-id lookup below).
+          found=false
+          for attempt in 1 2 3 4 5; do
```
tmathern, line 86: One more and you will be here for a minute :)

tmathern (approved):

PR #222, opened 2026-09-16
feat: Add credential holder for CAWG to c2pa-web.

cdmurph32:
Modify c2pa-node to use c2pa_cbor.


packages/c2pa-wasm/src/wasm_builder.rs
```diff
@@ -445,34 +446,54 @@ impl WasmBuilder {
...
 
     /// Sign an asset using the provided SignerDefinition, format, and source Blob.
     /// Use this method to get both the manifest bytes and the bytes of the signed asset.
+    ///
+    /// See [`Self::sign`] for `identity_assertions`.
     #[wasm_bindgen(js_name = signAndGetManifestBytes)]
     pub async fn sign_and_get_manifest_bytes(
         &mut self,
         signer_definition: &SignerDefinition,
+        identity_assertions: JsValue,
```
ale-adobe, line 483: Consider adding this as the last positional argument to preserve backward positional compatibility, although overall this would be a breaking change either way.

.changeset/beige-webs-tan.md
```diff
@@ -0,0 +1,7 @@
+---
+'@contentauth/c2pa-node': patch
+'@contentauth/c2pa-wasm': patch
```
ale-adobe: Public function signature changes are breaking for `c2pa-wasm` at the `wasm-bindgen` layer, so this should be a minor version bump.

packages/c2pa-wasm/src/wasm_identity_assertion.rs
```diff
@@ -0,0 +1,281 @@
+// Copyright 2025 Adobe
```
ale-adobe: nit
```suggestion
// Copyright 2026 Adobe
```

packages/c2pa-wasm/src/wasm_identity_assertion.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    if let Some(assertion_size) = size {
+        if assertion_cbor.len() > assertion_size {
+            return Err(c2pa::Error::BadParam(format!(
+                "Serialized assertion is {len} bytes, which exceeds the planned size of {assertion_size} bytes",
+                len = assertion_cbor.len()
+            )));
+        }
+
+        ia.pad1 = vec![0u8; assertion_size - assertion_cbor.len() - 15];
```
ale-adobe, line 184: Not super familiar with Rust, but any chance this could underflow? I see this was copied over from the Neon code, so it's existing code already.

packages/c2pa-wasm/src/wasm_identity_assertion.rs
```diff
@@ -0,0 +1,281 @@
...
+
+        assertion_cbor.clear();
+        c2pa_cbor::to_writer(&mut assertion_cbor, &ia)
+            .map_err(|e| c2pa::Error::BadParam(e.to_string()))?;
+
+        ia.pad2 = Some(ByteBuf::from(vec![
+            0u8;
+            assertion_size - assertion_cbor.len() - 6
+        ]));
```
ale-adobe, line 193: Similar question here.

packages/c2pa-wasm/src/wasm_credential_holder.rs
```diff
@@ -0,0 +1,165 @@
+// Copyright 2025 Adobe
```
ale-adobe: nit
```suggestion
// Copyright 2026 Adobe
```

packages/c2pa-web/src/lib/builder.spec.ts
```diff
@@ -1128,6 +1129,73 @@ describe('builder', () => {
         expect(manifestStore.active_manifest).toBeDefined();
       });
     });
+
+    describe('sign with a CAWG identity assertion', () => {
```
ale-adobe, line 1133: Can we have multiple CAWG identity assertions? How does that work?

packages/c2pa-wasm/src/wasm_identity_assertion.rs
tmathern: I am more curious as to where the `15` comes from (define it in a var or constant?).

packages/c2pa-wasm/src/wasm_identity_assertion.rs
tmathern: I am more curious as to where the `6` comes from (define it in a var or constant?).

packages/c2pa-wasm/src/wasm_credential_holder.rs
```diff
@@ -0,0 +1,165 @@
...
+// reason, wasm-bindgen appears to mishandle JsErrors when created in a Firefox web worker.
+//
+// See: https://github.com/wasm-bindgen/wasm-bindgen/issues/4961
+
+impl WasmCredentialHolder {
+    /// Build a `WasmCredentialHolder` from a JS object with `sigType`,
+    /// `reserveSize`, and `sign` fields (a subset of the identity-assertion
+    /// definition object; see `WasmIdentityAssertionBuilder::from_definition`).
+    pub(crate) fn from_definition(def: &JsValue) -> Result<Self, JsString> {
+        let reserve_size: Number = Reflect::get(def, &"reserveSize".into())?.into();
```
tmathern: Since coming from js, do you need to check it's not null undefined, NaN etc? Or is this handled OOTB here?
ps: `f64` may also be negative, it would be weird, but somewhere should be checked.

packages/c2pa-web/src/lib/signer.ts
```diff
@@ -36,3 +36,99 @@ export async function getSerializablePayload(
...
+   * always included).
+   */
+  referencedAssertions?: string[];
+  /** Named actor roles to attach to this identity assertion. */
+  roles?: string[];
+}
+
+/** Options accepted by `Builder.sign`/`Builder.signAndGetManifestBytes`. */
+export interface SignOptions {
+  identityAssertions?: IdentityAssertion[];
```
tmathern, line 96: For multiple assertions I think you will need current main/work-in-progress v0.91.0.

In current stable, I see this for dynamic assertions: https://github.com/contentauth/c2pa-rs/blob/c2pa-v0.90.22/sdk/src/store.rs#L2577-L2585 (I think it means the array wouldn't work here).

On v0.91.0 release candidate, it says this: https://github.com/contentauth/c2pa-rs/blob/c2pa-rc-v0.91.0-rc.1/sdk/src/store.rs#L2656-L2671

Probably not blocking, but goog to know since things are in-flight at time of writing.

packages/c2pa-web/src/lib/builder.spec.ts
tmathern: "It depends" it seems. (In theory it is possible, I think the spec allows it).
See https://github.com/contentauth/c2pa-js/pull/222#discussion_r4032764519

tmathern (approved):
(Some comments I had may be due to code similar to c2pa-rs, and therefore might be relevant there too).

.changeset/beige-webs-tan.md
```diff
@@ -0,0 +1,7 @@
...
+'@contentauth/c2pa-web': patch
+---
+
+Add credential holder for CAWG to c2pa-web.
```
ale-adobe, line 7: Let's also note the change for `c2pa-node`, which is the change to use `c2pa_cbor`.

packages/c2pa-web/src/lib/builder.spec.ts
ale-adobe: In that case, it would be nice to have a test case just verifying that we can handle multiple CAWG identity assertions too.

packages/c2pa-web/src/lib/signer.ts
```diff
@@ -36,3 +36,99 @@ export async function getSerializablePayload(
...
+  url: string;
+  hash: Uint8Array<ArrayBuffer>;
+  alg?: string;
+}
+
+/**
+ * The data a CAWG {@link CredentialHolder} is asked to sign over. Passed to
+ * {@link CredentialHolder.sign} exactly as built by c2pa-wasm's
+ * `WasmCredentialHolder` (camelCase field names, matching the
+ * adobe-web/utilities `encodeSignerPayload` contract).
```
ale-adobe: We shouldn't mention the internal libraries here.

ale-adobe (changes requested):
Couple of small nits and requests. See previous review as well.

.changeset/beige-webs-tan.md
cdmurph32: Makes sense. I don't know if anyone uses c2pa-wasm by itself, but if we have a version number, we might as well use it.

packages/c2pa-wasm/src/wasm_credential_holder.rs
cdmurph32: I'm going to add some checks there. I think you are right.

packages/c2pa-wasm/src/wasm_builder.rs
cdmurph32: I think it makes more sense from a code reading perspective to have it here, since it mirrors c2pa-rs. I don't think anyone uses c2pa-wasm directly, so readability matters more.

packages/c2pa-wasm/src/wasm_identity_assertion.rs
cdmurph32: I copied it from https://github.com/contentauth/c2pa-rs/blob/main/sdk/src/identity/builder/identity_assertion_builder.rs#L297
I'll add a comment. Unfortunately I don't have the time to make that public.

packages/c2pa-wasm/src/wasm_identity_assertion.rs
cdmurph32: Please see my response to the previous comment.

packages/c2pa-web/src/lib/signer.ts
cdmurph32: I'll discuss this with you and Austin so we are ready when c2pa-rs 0.91 lands.

packages/c2pa-web/src/lib/builder.spec.ts
cdmurph32: Please see my response to @tmathern 's comment.

tmathern: Yes, it will depend on the underlying c2pa-rs version. I pointed it out so you know there is a discrepancy depending on which underlying c2pa-rs you have. Currently v0.90.something, aka current stable as time of writing, does not seem to support it.

tmathern (approved):

ale-adobe (approved):
LGTM, but reminder to please update the year on those copyright headers before you merge. I don't think they made it into your last commit with all the other changes.

PR #224, opened 2026-09-18
fix: Backoff retries on some workflows

tmathern:
A recent change already retried the npm repository presence check before publishing the binary (See https://github.com/contentauth/c2pa-js/pull/220).

It seems the wait delay is not long enough, so switching to a backoff approach (to avoid hammering the registry), letting retries run for approximately 3 minutes.

Also caps the job to 1hour, in case it would "really" hang. 1 hour should be generous enough for the job to finish a normal run.


cdmurph32 (approved):
approved

